PR #70, opened 2014-10-13
Added header files so they are visible in VS solution tree.

larshg:
Not necessary for compiling, but is nice for making changes.


floe:
I'll merge this; but could you add the remaining headers (e.g. registration.h) too, please?

larshg:
@laborer2008  it cannot be done, since test_opengl_depth_packet_processor.cpp contains a main method and gives linking problems in the libfreenect2 library.

PR #71, opened 2014-10-15
Add timestamps to Frame data.

hovren:
Timestamps for the color data is extracted from a footer that
begins after the JPEG End-Of-Image (0xFF 0xD9)  marker.
The timestamp for the frame is located  36 bytes from the end of the data.


hovren:
I know the RGB footer PRs still have not been merged, but could we possibly merge this PR anyway (after rebasing on master)?

This patch is quite small, and as soon as the RGB footer parsing is solved, just one line of code is required to fetch the timestamp from the footer struct instead of from the current hard coded buffer offset.

xlz:

PR #72, opened 2014-10-15
Robust check for arrived color frame.

hovren:
Previous implementation assumed a full frame was reached if the current transfer is
smaller than 0x4000.
This implementation instead sees if the data footer has arrived by checking the expected
length against the total buffer size, and that the sequence number agrees.
This should be more robust.


xlz:

PR #73, opened 2014-10-16
Rgb packet handling

larshg:
A try at getting a better rgb packet handling with inspiration from hovren's pull request: https://github.com/OpenKinect/libfreenect2/pull/72

Its still a work in progress, but I wouldn't mind any feedback on the way :+1:

### DIFF
--- a/examples/protonect/src/rgb_packet_stream_parser.cpp
+++ b/examples/protonect/src/rgb_packet_stream_parser.cpp
@@ -53,7 +53,8 @@ LIBFREENECT2_PACK(struct RgbPacketFooter {
     float gain; // ranges from 1.0 when camera is clear to 1.5 when camera is covered.
     uint32_t magic_footer; // is 'BBBB' equal 0x42424242
     uint32_t packet_size;
-    uint8_t unknown4[16]; // seems to be 0 all the time.
+    float unknown4; // consistenly 1.0
+    uint8_t unknown5[12]; // 0 all the time.
 });
 
 RgbPacketStreamParser::RgbPacketStreamParser() :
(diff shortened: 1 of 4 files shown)


examples/protonect/src/rgb_packet_stream_parser.cpp
```diff
@@ -36,11 +37,26 @@ namespace libfreenect2
 LIBFREENECT2_PACK(struct RawRgbPacket
 {
   uint32_t sequence;
-  uint32_t unknown0;
+  uint32_t magicHeader; // is 'BBBB' equal 0x42424242
```
christiankerl: rename to `magic_header`

examples/protonect/src/rgb_packet_stream_parser.cpp
```diff
@@ -36,11 +37,26 @@ namespace libfreenect2
...
 
   unsigned char jpeg_buffer[0];
 });
 
+struct RgbPacketFooter {
+    uint32_t magicHeader; // is '9999' equal 0x39393939
```
christiankerl: rename to `magic_header`

examples/protonect/src/rgb_packet_stream_parser.cpp
```diff
@@ -36,11 +37,26 @@ namespace libfreenect2
...
+    uint32_t sequence;
+    int32_t unknown1; // varies from 0 to 60 with steps of 4
+    int32_t unknown2; // seems 0 always
+    int32_t unknown3; // seems 0 always
+    uint32_t timestamp;
+    float exposure; // ? ranges from 0.5 to about 60.0 with powerfull light at camera or totally covered
+    float gain; // ? ranges from 1.0 when camera is clear to 1.5 when camera is covered.
+    uint32_t magicFooter; // is 'BBBB' equal 0x42424242
```
christiankerl: rename to `magic_footer`

examples/protonect/src/rgb_packet_stream_parser.cpp
```diff
@@ -36,11 +37,26 @@ namespace libfreenect2
...
+    uint32_t packet_size;
+    float version; // ? this seems to be consistent 1.0
```
christiankerl: why do you call this version? `unknown4` would be fine

examples/protonect/src/rgb_packet_stream_parser.cpp
```diff
@@ -58,49 +74,65 @@ void RgbPacketStreamParser::setPacketProcessor(BaseRgbPacketProcessor *processor
...
-    // not full transfer buffer and we already have some data -> signals end of rgb image packet
-    // TODO: better method, is unknown0 a magic? detect JPEG magic?
-    if(length < 0x4000 && fb.length > sizeof(RgbPacket))
-    {
-      // can the processor handle the next image?
-      if(processor_->ready())
-      {
-        buffer_.swap();
-        Buffer &bb = buffer_.back();
+      buffer_.swap();
```
christiankerl: the buffers should only be swapped if the processor is ready otherwise it might still operate on the last back buffer, which you turn here into our new front buffer (to which we will write at the next call of this method)

examples/protonect/src/rgb_packet_stream_parser.cpp
```diff
@@ -36,11 +37,26 @@ namespace libfreenect2
 LIBFREENECT2_PACK(struct RawRgbPacket
 {
   uint32_t sequence;
-  uint32_t unknown0;
+  uint32_t magicHeader; // is 'BBBB' equal 0x42424242
 
   unsigned char jpeg_buffer[0];
 });
 
+struct RgbPacketFooter {
```
christiankerl: this struct has to be wrapped with LIBFREENECT2_PACK()

xlz:
First, can you please squash these into a single commit? The second one is cosmetic.

As a matter of fact, a custom hardware accelerated JPEG decoder I'm testing cannot handle garbage after JPEG. So this extra check is actually required for compatibility.

My experiments show your `unknown1` is the length of a filler before the footer. Some new specifications:
```c
// starting from JPEG EOI: 0xff 0xd9
// char _pad_0xa5[]; //0-3 bytes alignment of 0xa5
// char _filler[_filler_length] = "ZZZZ...";
LIBFREENECT2_PACK(struct RgbPacketFooter {
    uint32_t magic_header; // is '9999' equal 0x39393939
    uint32_t sequence;
    int32_t unknown1; // _filler_length
[...]
```

Also, `version` is shown to be always 0x3f800000 (1.0f). It's likely a float number of certain meaning, but it is probably not representing a version number. The rest of the `unknown*` fields are also shown to be zeros in all my experiments. Maybe you can rename the `unknown*` fields to be `reserved*` because they are not exactly unknown.

I wrote and tested some code to find out the exact size of the JPEG image by searching for its EOI.  [For your reference](https://github.com/xlz/libfreenect2/blob/9ee5e21f754e0fbae99b16b9cb21696fb46d93d4/examples/protonect/src/tegra_jpeg_rgb_packet_processor.cpp#L237-L271). Following the new specifications the code will finish in finite steps and if EOI is not found, returning failure is justified. In my testing it always worked.

christiankerl:
@xlz would it suffice to compute the correct jpeg_buffer_length using the filler_length?

@larshg can you rename unknown1 accordingly?

examples/protonect/src/rgb_packet_stream_parser.cpp
```diff
@@ -58,49 +73,66 @@ void RgbPacketStreamParser::setPacketProcessor(BaseRgbPacketProcessor *processor
...
   {
-    if(fb.length + length <= fb.capacity)
-    {
-      memcpy(fb.data + fb.length, buffer, length);
+    RgbPacketFooter* footer = reinterpret_cast<RgbPacketFooter *>(&buffer[i]);
+
+    // if magic markers match
+    if (length - i >= sizeof(RgbPacketFooter) && footer->magic_header == 0x39393939 && footer->magic_footer == 0x42424242)
+    {  
+      memcpy(&fb.data[fb.length], buffer, length);
```
christiankerl: This copies all remaining data to the fb. I guess it would be more correct to just copy i+sizeof(footer) and copy any remaining data to the new fb after the swap?

xlz:
@christiankerl Yes.

[Here is my take](https://github.com/xlz/libfreenect2/commit/8296e0650488eca773b64deab99a73df99e82930) to compute the exact value of jpeg_buffer_length using filler_length.

Several changes:
* Return immediate after buffer overflow because the next JPEG packet will be incomplete anyway.
* Do not scan through every single byte checking for magic bytes. This is much more costly than a memcpy. The current implementation and the lack of reports about buffer overflow and image corruption show it is safe to assume the footer arrives at the very end of a data transfer, and there is no more data after the footer within a data transfer.

xlz:
[My take v2](https://github.com/xlz/libfreenect2/commit/1bbfc643279b47bee6a035fbb64423ac8f7cf84d).

Changes from v1:
* Corrected size calculation about bytes before JPEG
* Tested on Linux

This is ready for cherry picking.

larshg:
Nice suggestions for improvements @xlz. I have tried to incorporate most of them - Still lacking to calculate the exact jpeg buffer size with the 0-3 padding bytes, but I'll try look at it tomorrow.

I have removed the for loop, but this requires the assumption that there will be no concatenated images in the USB transfers - which I'm not qualified to determine can or will not happen?

Over 1000 runs in the onDataReceived method, I tried with and without a for loop and I measured difference from 160 ms to 1.6 ms in processing time - so indeed it is more costly, but difference compared to the overall processing is minimal.
But if there is no reason to do it (more secure, removes doubt of the assumption of concatenated images), we might as well remove it?

xlz:
In fact you can totally detect the magic bytes in the middle of an image. There is no guarantee these 8 bytes won't appear. If our assumption does not hold, report of buffer overflows will immediately show up.

(It is similar for depth processing which is also doing scanning currently. I'll put on my patch shortly.)

On the side of performance, on ARM platform the CPU is significantly weaker for these kind of memory scanning.

xlz:
What's the status of these two stream processor PR #73 #75 ?

larshg:
Closing in favor of #221 .

PR #128, opened 2015-01-12
fix for wrong define name: WITH_OPENCL_SUPPORT -> LIBFREENECT2_WITH_OPENCL_SUPPORT

wiedemeyer:
Package pipeline not building correctly because the the define name was not updated.


christiankerl:
whoops, thanks!

PR #148, opened 2015-01-31
Added sequence in Frame

MasWag:
This commit enables to get sequence from Frame.


xlz:

PR #151, opened 2015-02-02
remove OpenCV dependency

christiankerl:
this removes the OpenCV dependency by replacing cv::Mat with a lightweight implementation in CpuDepthPacketProcessor.

By removing OpenCV we loose:
 - timing of the packet processors 
 - display of images in Protonect

However, for displaying the images I think about a simple viewer implementation using OpenGL (if available). What about timing, should we use std::chrono if compiled with C++11 support and otherwise drop it? 

NOTE: this is work in progress


floe:
Great, thanks! (Had some half-baked code with the same goal floating around, you beat me to it :-)
Since we require some pretty recent libraries anyway, I think it might also be worth simply requiring C++11 and not mess around with backwards compatibility. After all, the standard is already over 3 years old, and g++/VC++ support all required bits.

xlz:
Visual Studio has horrible std::chrono implementation ([high_resolution_clock with 1ms resolution](http://stackoverflow.com/questions/16299029/resolution-of-stdchronohigh-resolution-clock-doesnt-correspond-to-measureme)), and it's probably not getting better with MinGW because the timekeeping infrastructure of Windows generally sucks. It's probably OK to just steal [OpenCV's getTickCount](https://github.com/Itseez/opencv/blob/7854ef3840b8216396b6138020404ea1e207d671/modules/core/src/system.cpp#L395-L435) and there would be no license issue.

floe:
@xlz good point about getTickCount - however, since it's just used for rough performance estimates, I think 1 ms resolution might be enough?

xlz:
@floe 1ms error means ~10Hz error. But it's only for Windows users, I guess they can use the official sdk.

larshg:
@xlz how does 1ms equals 10 Hz error? Shouldn't 1 ms be equal to 0.001 Hz? Since you need 1000 ms for a second?

Also, the resolution should have been corrected in the VS2014(2015) CTP2, mentioned in a comment of the thread you link to :) - For some reason I cannot find the bug on their bug tracker though...

And as @floe points out its only for rough processing times averaged over 100 frames, it really doesn't matter?

xlz:
@larshg Taking an example of what I have been measuring, 1000ms/8ms = 125Hz, 1000ms/9ms = 111Hz. That's 14Hz error in the rate of frame per second.

xlz:
I have looked at how to implement an OpenGL viewer and it is not easy. For 0.1 release it is easier to just split the dependency of OpenCV out from libfreenect2, but let `Protonect` continue depending OpenCV as a standalone app/example/test case.

On C++11, I am still wary about breaking this backward compatibility. C++11 has ABI incompatibility with C++98. There will be issues to build libfreenect2 with C++11 and link it with C++98 objects. See https://gcc.gnu.org/wiki/Cxx11AbiCompatibility .

floe:
Merge conflict; closing in favor of #261 (which includes the changes proposed here anyway).

PR #166, opened 2015-02-12
Remove the VS solution as it is outdated.

larshg:
Updated README for now.

I might look into a more detailed explanation for dependencies and the like.


floe:
Can you currently use CMake on Windows to generate a VS solution?

larshg:
You can yes - but it doesn't find all the dependencies and mine look like this:
master

This should change with #68 merged. As for now, you can insert all the paths/libs manually.

But if I use the so called VS solution that already exist - this is what I have:
mastervssolution

Which doesn't eveny build the libfreenect2 project and hence, the protonect project won't work at all. And all the paths in the solution is hardcoded and doens't match mine etc.

floe:
As nobody has touched the VS solution since, it's even more outdated by now, so let's get rid of it. Merging.

PR #167, opened 2015-02-13
Make opencl processor compiling on newer linux nvidia CL version

goldhoorn:


goldhoorn:
For more information see http://stackoverflow.com/questions/19405596/cannot-compile-opencl-application-using-1-2-headers-in-1-1-version

wiedemeyer:
Hi,
I just installed a 14.04.1 on a system with a nvida GPU yesterday. I installed the nvidia 346.35 driver from ppa:xorg-edgers and the 'opencl-headers'. And libfreenect2 compiled with OpenCL support and was running without any errors related to mismatching OpenCL versions.

lolz0r:
I have seen this same error and fixed it in a similar fashion. I am running ubuntu 14.04 nvidia driver 331; all straight from the standard sources.

floe:
@wiedemeyer @lolz0r so this can be merged without ill effect?

wiedemeyer:
I just tested the modifications on Nvidia and AMD, and it works without problems. So you can merge it safely.
I still don't really like the solution, because it is only needed for Nvidia cards with the proprietary drivers from the Ubuntu repos. With more recent drivers or GPUs from Intel and AMD it works fine without forcing it to use the deprecated API. The optimal solution would be to just add these lines, when a Nvidia card with driver version lower than `x` is used. Or to set the build options for the OpenCL code to `"-cl-std=CL1.1"` if a Nvidia card with those drivers is detected at runtime.

floe:
Is there a #define or something to detect the version of the Nvidia driver? You're probably right that using deprecated APIs by default is probably not such a good idea, they're usually deprecated for a reason...

wiedemeyer:
I am not sure if there are any defines. Maybe it is possible to get the information within CMake, for example `nvidia-settings -v` contains the driver version in the output.

goldhoorn:
This would make this even more ugly.

Is there are really any problem for merging as it is for other nvidia 
versions or AMD cards. Or do we discussing some theoretical problems 
that might occur?

floe:
Good point, merged.

wiedemeyer:
On one computer where CUDA is installed, compiling libfreenect2 failed with the following error:
```
In member function ‘void libfreenect2::OpenCLDepthPacketProcessorImpl::listDevice(std::vector<cl::Device>&)’:
libfreenect2/examples/protonect/src/opencl_depth_packet_processor.cpp:271:12: error:
‘CL_DEVICE_TYPE_CUSTOM’ was not declared in this scope
```
After removing the 3 lines from this pull-request it compiled without issues.

xlz:
I am revisiting this issue.

The root cause is the version of OpenCL ICD **loader** (libOpenCL.so) must match that of OpenCL headers (cl.h, not cl.hpp). If 1.2 cl.h provides symbols that do not exist in 1.1 libOpenCL.so, then linking fails. But this has nothing to do with OpenCL driver version.

There are many ways to solve this without changing a line of code.

* Use CUDA Toolkit 7.5 (1.2 libOpenCL.so and headers)
* Use any older CUDA Toolkit (1.1 libOpenCL.so and headers)
* Use ocl-icd-libopencl1 (1.2 libOpenCL.so), opencl-headers (1.2), with 1.1 Nvidia driver

The only way that doesn't work is

* Use nvidia-libopencl1 (1.1 libOpenCL.so), opencl-headers (1.2)

This can be trivially fixed by `apt-get install ocl-icd-libopencl1`.

This problem with this PR is it forces everyone else (anyone with correct setup on Linux, everyone on Windows and Mac OSX) to not use OpenCL 1.2 APIs, and turn off new fixes in OpenCL 1.2 that are not backported to 1.1.

I will create a compile time check to match libOpenCL.so version with headers. If not matching, issue a warning and use compatibility as in this PR.

PR #177, opened 2015-03-05
Example code to stream video feeds to v4l using v4l2loopback

jfrey-xx:
Hello,

For my personal use I had to acquire kinect2 cameras using V4L. I coded a new example that dumps the frame to a fake /dev/video* devices using v4l2loopback.

It's a quick hack, not optimized and that could be outdated soon since #151 propose to get rid of opencv, but it works (tested  with ubuntu 14.04). The video feeds can now be viewed -- and recorded -- with guvcview or VLC -- may have to select explicitly Y16 as video format input for IR and depth feeds.

It more of a proof of concept, but if one thinks that the code is of interest I'm eager to improve it -- starting with video devices *not* hard-coded. Comments welcomed :)


floe:
This is a great idea, but would it be possible to implement this as a standalone program on top of libfreenect2? Right now, the pull request introduces a second modified copy of the whole source code, which would be entirely unmaintainable...

jfrey-xx:
You're right, there *some* code duplication. I'll clean up and start a fresh repository to facilitate the development. Thanks for the feedback!

Issue #41, opened 2014-08-15
Depth-color registration parameters are unknown


floe:
I'm still trying to get the calibration/registration parameters sorted out, but I'm kinda stuck, so I'll try to write down what I think I know and hope that someone can provide some additional pointers... 

The intrinsics for the depth camera seem pretty certain - for the two
devices I own, they are:

    depth camera intrinsic parameters:
    fx 368.096588, fy 368.096588, cx 261.696594, cy 202.522202
    depth camera radial distortion coeffs:
    k1 0.084061, k2 -0.271582, p1 0.000000, p2 0.000000, k3 0.101907

    depth camera intrinsic parameters:
    fx 365.402802, fy 365.402802, cx 260.925507, cy 205.594604
    depth camera radial distortion coeffs:
    k1 0.095575, k2 -0.277055, p1 0.000000, p2 0.000000, k3 0.097539

This is consistent with a camera having square pixels at a resolution of 512*424 (although cy seems a bit off-center to me).

Next are color camera intrinsics - there's a 25-value block which contains the following 5 initial values on both cameras:

    1081.37 - might again be focal length in pixels - FOV?
     959.5  - exactly half the x resolution -> optical center x?
     539.5  - exactly half the y resolution -> optical center y?
     863    - not really a clue - perhaps x shift relative to depth cam?
      52    - perhaps y shift relative to depth cam?

About the 20 other values, I have no real clue, although part of them look like they might form a 4x4 transformation matrix... 

Any additional comments/ideas?

sh0:
I have some good news, I believe I have figured out the depth-color parameters.

The important parameters reside in `intrinsics[25]` (`RgbCameraParamsResponse` structure). New layout is as follows:

    struct s_intrinsics {
        float color_f;
        float color_cx;
        float color_cy;
        
        float shift_d;
        float shift_m;
        
        float mx_x3y0; // xxx // 0x40
        float mx_x0y3; // yyy // 0x44
        float mx_x2y1; // xxy // 0x48
        float mx_x1y2; // yyx // 0x4c
        float mx_x2y0; // xx  // 0x50
        float mx_x0y2; // yy  // 0x54
        float mx_x1y1; // xy  // 0x58
        float mx_x1y0; // x   // 0x5c
        float mx_x0y1; // y   // 0x60
        float mx_x0y0; //     // 0x64
        
        float my_x3y0; // xxx // 0x68
        float my_x0y3; // yyy // 0x6c
[...]

PR #190, opened 2015-03-25
add basic Registration class

floe:
I've announced my intent to write this a while ago in #41, now I've finally gotten around to an initial implementation. Thanks again to @sh0 for the hard work and M. Vasileiadis for testing the unfinished code.

### DIFF
--- a/examples/protonect/CMakeLists.txt
+++ b/examples/protonect/CMakeLists.txt
@@ -87,6 +87,7 @@ SET(SOURCES
   
   src/usb_control.cpp
   src/command_transaction.cpp
+  src/registration.cpp
   src/libfreenect2.cpp
   ${LIBFREENECT2_THREADING_SOURCE}
   ${RESOURCES_INC_FILE}
--- a/examples/protonect/include/libfreenect2/libfreenect2.hpp
+++ b/examples/protonect/include/libfreenect2/libfreenect2.hpp
@@ -45,6 +45,30 @@ public:
   struct ColorCameraParams
   {
     float fx, fy, cx, cy;
+
+    float shift_d, shift_m;
+
+    float mx_x3y0; // xxx
+    float mx_x0y3; // yyy
+    float mx_x2y1; // xxy
+    float mx_x1y2; // yyx
+    float mx_x2y0; // xx
+    float mx_x0y2; // yy
+    float mx_x1y1; // xy
+    float mx_x1y0; // x
+    float mx_x0y1; // y
+    float mx_x0y0; // 1
+
+    float my_x3y0; // xxx
+    float my_x0y3; // yyy
+    float my_x2y1; // xxy
+    float my_x1y2; // yyx
+    float my_x2y0; // xx
+    float my_x0y2; // yy
+    float my_x1y1; // xy
+    float my_x1y0; // x
+    float my_x0y1; // y
+    float my_x0y0; // 1
   };
 
   struct IrCameraParams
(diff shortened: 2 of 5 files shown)


floe:
@christiankerl @wiedemeyer I will probably merge this soon-ish into master, as it doesn't currently affect any existing code. For future PRs which actually use the Registration class, I will wait for your input.

larshg:
Hi @floe 

I tried out the registration class - you can see my changes here (There are also some other changes, but nevermind those):

https://github.com/larshg/libfreenect2/tree/registration

Example here:
image

There are some artifacts when there is two objects in the view, making a subshadow of the closest object - I'm not sure why this is happening...

Is the intention that the registration class should be used "outside" libfreenect2 or should it be used inside, as to create a ColoredDepthFrame producer, which you can add a listener to like with the RGB, ir and Depth?

Are there any reason why you do the undistort_depth in two for loops? and is it the intention that the undistort map should be used somewhere else?

To be honest, I'm not that familiar with the multiple view geometry - so I'll try to read it up a bit :)

floe:
Thanks for testing! I don't quite understand what you mean by "subshadow", screenshot looks correct to me?

I would indeed like to create another class of FrameListener, and hopefully take advantage of OpenGL/CL acceleration, too. However, I'm currently a bit swamped with work, so this might take a while... for the moment, it's therefore probably easiest to just merge most of your test code into the apply() method and give people the opportunity to call it on their own.

Calling undistort_depth twice is useless, you're right. However, I'd like to keep the undistortion table separately and add a method to _only_ perform undistortion without depth-color mapping.

lolz0r:
I pulled and built from larshg's fork (https://github.com/larshg/libfreenect2/tree/registration). Cool stuff! However, I still see misalignment as seen in this screenshot:
misalignment

This being said I see similar artifacting when using the official kinect SDK. This leads me to believe that some kind of damage / error has occurred to my kinect hardware causing it to deviate from the factory calibrated values.

Is there anyway to recalibrate the values needed for the registration system using something like: (https://github.com/code-iai/iai_kinect2/tree/master/kinect2_calibration) ?

I went ahead and built the iai_kinect2 calibration tool and was able to generate calibration values. It makes YAML files with contents like:

%YAML:1.0
cameraMatrix: !!opencv-matrix
   rows: 3
   cols: 3
   dt: d
   data: [ 1.0678176639391463e+03, 0., 9.8017834376380301e+02, 0.,
       1.0680561237549337e+03, 5.5693821427066280e+02, 0., 0., 1. ]
distortionCoefficients: !!opencv-matrix
   rows: 1
   cols: 5
   dt: d
   data: [ 4.0403537419643677e-02, -1.8850131673369398e-02,
       2.2666058394713523e-03, 2.1452074250293118e-03,
       -2.0779845401099672e-02 ]
rotation: !!opencv-matrix
   rows: 3
   cols: 3
   dt: d
   data: [ 1., 0., 0., 0., 1., 0., 0., 0., 1. ]
projection: !!opencv-matrix
   rows: 4
   cols: 4
   dt: d
   data: [ 1.0678176639391463e+03, 0., 9.8017834376380301e+02, 0., 0.,
       1.0680561237549337e+03, 5.5693821427066280e+02, 0., 0., 0., 1.,
       0., 0., 0., 0., 1. ]

...
How do I take these values and utilize them in lieu of the factory calibrated values?

Thank you for any help you may offer!

floe:
I think that's not an issue with your hardware, that's a fundamental geometric limitation. I remember a similar discussion in libfreenect1: https://github.com/OpenKinect/libfreenect/pull/347 
Since there are some pixels that only the depth camera can see, but not the color camera and vice versa, there will always be some pixel duplication unless you apply a depth filter.

xlz:
Based on @larshg's implementation I added the depth filter to remove depth pixels that can't be seen by the RGB camera

2015-05-06-132923_1920x1080_scrot

The code is here https://github.com/xlz/libfreenect2/tree/registration

Issue #157, opened 2015-02-07
Fails to build from source on ARM introduced in #103


xlz:
```
Linking CXX shared library ../lib/libfreenect2.so
CMakeFiles/freenect2.dir/src/rgb_packet_stream_parser.cpp.o:(.rodata+0x0): multiple definition of `typeinfo name for libfreenect2::PacketProcessor<libfreenect2::RgbPacket>'
CMakeFiles/freenect2.dir/src/packet_pipeline.cpp.o:(.rodata+0x0): first defined here
CMakeFiles/freenect2.dir/src/rgb_packet_stream_parser.cpp.o:(.data.rel.ro+0x0): multiple definition of `typeinfo for libfreenect2::PacketProcessor<libfreenect2::RgbPacket>'
CMakeFiles/freenect2.dir/src/packet_pipeline.cpp.o:(.data.rel.ro+0x0): first defined here
[...]
collect2: error: ld returned 1 exit status
```

This bug is introduced in #103. [Here](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=608029) is a better background for this bug. Basically, these two explicit instantiations
```cpp
// explicit instantiation and export to make vsc++ happy
template class LIBFREENECT2_API PacketProcessor<DepthPacket>; 

// explicit instantiation and export to make vsc++ happy
template class LIBFREENECT2_API PacketProcessor<RgbPacket>;
```
in headers being included multiple times violates C++ standard 14.7 clause 5:

 > an explicit instantiation definition shall appear at most once in a program

and GCC on ARM fails because of optimizations assuming the standard.

christiankerl:
so what do you suggest? Is wrapping the explicit instantiations in an ifdef for msvc an option?

xlz:
@christiankerl Yes we may need a msvc specific flag but I don't have msvc to test it.
@larshg Can you look into this?

emrainey:
I would recommend
 #if defined(_MSC_VER)  
 #endif
 From https://msdn.microsoft.com/en-us/library/b0084kay.aspx

PR #207, opened 2015-04-15
Fix FTBFS on ARM introduced in PR #103

xlz:
PR #103 tried to fix a linking issue in Visual Studio 2013 on
Windows 7. It added multiple explicit template instantiations
which violates the standard and results in failure to build
from source on ARM.

Further testing failed to reproduce the linking issue with
Visual Studio 2013 on Windows 8.1. Thus this commit removes
the explicit template instantiations.

This fixes issue #157.

### DIFF
--- a/examples/protonect/include/libfreenect2/depth_packet_processor.h
+++ b/examples/protonect/include/libfreenect2/depth_packet_processor.h
@@ -44,8 +44,6 @@ struct LIBFREENECT2_API DepthPacket
   size_t buffer_length;
 };
 
-// explicit instantiation and export to make vsc++ happy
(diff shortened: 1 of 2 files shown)


rammstan:
Hello, wich plateform/computer arm do you use ?

xlz:
Jetson TK1

larshg:
I just tried to reproduce the error again, but for some reason its not causing errors (Win 7 VS 2013). Maybe due to that I started using the community edition of VS2013 - and the error only happen when using VS express.... Or perhaps an updated version of VS2013.

Feel free to merge this one @christiankerl or @floe

floe:
Tested this with gcc, just to be safe, and it still works, so I'm merging this.

PR #208, opened 2015-04-15
Update build instructions for Windows with MSVC and OpenCL

xlz:


floe:
I'd like to merge the README changes separately, as the changes in CMakeLists.txt could possibly collide with #68. I'm also changing the README in #240, so I'd suggest the following: I'll merge #240 first, and you submit a new PR on top of that with your build instructions - agreed?

floe:
Apparently git[hub] is smart enough to recognize that the README changes don't actually conflict. Nice. Still, would be great if you could split this into two.

xlz:
#68 will change the build workflow quite a lot. This PR can be subsumed by 68.

Issue #185, opened 2015-03-23
Segfault 11 (on mac)


julienaubert:
Regardless if I use OpenCL, CPU, or OpenGL.

[OpenCLDepthPacketProcessor] avg. time: 19.8993ms -> ~50.2531Hz
q[Freenect2DeviceImpl] stopping...
[Freenect2DeviceImpl] disabling usb transfer submission...
[Freenect2DeviceImpl] canceling usb transfers...
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
[TransferPool::submit] transfer submission disabled!
libusb: error [darwin_cancel_transfer] unknown endpoint type 49
libusb: error [libusb_cancel_transfer] cancel transfer failed error -2
Segmentation fault: 11

floe:
Does this happen when you quit, or randomly at some point during runtime?

julienaubert:
Each time when I quit.

floe:
OK, in this case, can you try to run Protonect from gdb and post a stacktrace from the segfault?

julienaubert:
[...]

PR #212, opened 2015-04-24
Handling race condition for pending_transfers_ and its iterators.

larshg:
I tried updating libusb to the newest master, which due to this commit:
https://github.com/libusb/libusb/commit/a886bb02c87dd5faf271fad595e4622f7027d347

Made a lot more errors on the transfer submitting than I experienced before. After investigating it a bit, I came to the conclusion that there is thread unsafe adding and removing going on, which occasionally made the iterators invalid, since the underlying list/queue had changed.

This pull request adds a mutex to allow only one thread at a time to access the pending_transfer_ list.
Can you verify this @christiankerl?

There might be more places to handle this problem, but for now I'm not experiencing the errors anymore.

I tested with the "older" version of libusb and this addition doesn't seem to break it, nor make any performance loss.

This might be related to the error in #185.

### DIFF
--- a/examples/protonect/include/libfreenect2/usb/transfer_pool.h
+++ b/examples/protonect/include/libfreenect2/usb/transfer_pool.h
@@ -29,6 +29,7 @@
 
 #include <deque>
 #include <libusb.h>
+#include <libfreenect2/threading.h>
 
 #include <libfreenect2/data_callback.h>
 
@@ -70,6 +71,8 @@ private:
   libusb_device_handle *device_handle_;
   unsigned char device_endpoint_;
 
+  mutex pending_transfers_lock_;
+  mutex idle_transfers_lock_;
   TransferQueue idle_transfers_, pending_transfers_;
   unsigned char *buffer_;
   size_t buffer_size_;
--- a/examples/protonect/src/transfer_pool.cpp
+++ b/examples/protonect/src/transfer_pool.cpp
@@ -60,10 +60,13 @@ void TransferPool::disableSubmission()
 
 void TransferPool::deallocate()
 {
+  idle_transfers_lock_.lock();
   for(TransferQueue::iterator it = idle_transfers_.begin(); it != idle_transfers_.end(); ++it)
   {
     libusb_free_transfer(*it);
   }
+  idle_transfers_lock_.unlock();
+
   idle_transfers_.clear();
 
   if(buffer_ != 0)
@@ -86,6 +89,7 @@ void TransferPool::submit(size_t num_parallel_transfers)
   {
     std::cerr << "[TransferPool::submit] too few idle transfers!" << std::endl;
   }
+  pending_transfers_lock_.lock();
 
   for(size_t i = 0; i < num_parallel_transfers; ++i)
   {
@@ -105,10 +109,14 @@ void TransferPool::submit(size_t num_parallel_transfers)
       std::cerr << "[TransferPool::submit] failed to submit transfer" << std::endl;
     }
   }
+
+  pending_transfers_lock_.unlock();
 }
 
 void TransferPool::cancel()
 {
+  pending_transfers_lock_.lock();
+  
   for(TransferQueue::iterator it = pending_transfers_.begin(); it != pending_transfers_.end(); ++it)
   {
     int r = libusb_cancel_transfer(*it);
@@ -118,6 +126,8 @@ void TransferPool::cancel()
       // TODO: error reporting
     }
   }
+
+  pending_transfers_lock_.unlock();
 
   //idle_transfers_.insert(idle_transfers_.end(), pending_transfers_.begin(), pending_transfers_.end());
 }
@@ -160,6 +170,7 @@ void TransferPool::onTransferCompleteStatic(libusb_transfer* transfer)
 
 void TransferPool::onTransferComplete(libusb_transfer* transfer)
 {
+  pending_transfers_lock_.lock();
   // remove transfer from pending queue - should be fast as it is somewhere at the front
   TransferQueue::iterator it = std::find(pending_transfers_.begin(), pending_transfers_.end(), transfer);
 
@@ -169,6 +180,8 @@ void TransferPool::onTransferComplete(libusb_transfer* transfer)
   }
 
   pending_transfers_.erase(it);
+
+  pending_transfers_lock_.unlock();
 
   // process data
   processTransfer(transfer);


xlz:
It seems the current libfreenect2 usage of libusb callback function just happens to not crash given that libusb has no guarantee the callback function will not be called from different threads. I think adding locks would be a ok temporary fix. 

For a better solution, it's better without introducing synchronization points here and there because that defeats the point of asynchronous transfers. This two queues `pending_transfers_` and `idle_transfers_`  basically juggle transfer structures around and the only point of the two queues is to determine whether a transfer is being worked on or being idle. The method of determining the status is to use thread unsafe iterators over the queues.

Putting the transfer status inside the transfer structure and using a std::vector to contain the transfers would also do the job in a thread-safe way.

xlz:
Bisection confirms a886bb0 is the commit that breaks the current libfreenect2.

Your patch did not fix the problem in my testing. I did not get any "failed to submit transfer" error, but the errors were a lot of ISO transfers being lost and depth packets being incomplete.

I also found `TransferPool::onTransferCompleteStatic` is always called from a same thread so there is no issue about thread safety.

One minor issue, your patch uses std::mutex and will fail to build if the user does not enable C++11.

larshg:
Hi @xlz 

The runtime error I get is that the iterator is invalid - nothing to do about failed to submit transfers :)

You might be right that its the same thread that executes onTransferCompleteStatic, but its the main thread that adds and removes transfers at startup and on close.

But as soon as one transfer is submitted, libusb spawns a thread and that one will eventually be using a iterator on the collections, while the main thread is adding or removing transfers too. This will make the iterator invalid.

Instead of using std::mutex I guess we can use a mutex from tinythread or another synchronization way.

xlz:
Then my error from a886bb0 is different from yours. I did not get any invalid iterator errors.

> You might be right that its the same thread that executes onTransferCompleteStatic, but its the main thread that adds and removes transfers at startup and on close.

What I did was printing the thread id at the position where you put the locks. Apparently `TransferPool::onTransferComplete` is the only entry point that accesses the iterators after initialization.

larshg:
I just figured that I can only get the error to happened when I run it in debug mode - probably because all the transfers get submitted before the first transfer is complete in release mode. Did you try in debug mode?

xlz:
Agreed. What you found is true even before a886bb0.

With the current [libusbx-winiso](https://github.com/JoshBlake/libusbx/tree/winiso), and with this line  removed in `transfer_pool.cpp`:
```
    std::cerr << "[TransferPool::submit] transfer submission disabled!" << std::endl;
```
I got "deque iterator not incrementable" error using Debug build, and hang-up using Release build when exiting Protonect.

Either you have to put locks everywhere, or have to consider reimplementing the transfer pool.

xlz:
@larshg 

It seems on Windows 886bb0 improves performance, but on Linux 886bb0 really breaks the original transfer_pool.cpp. 

Can you try https://github.com/xlz/libfreenect2/commit/c6f6e57df3671285812e5ae3ac9ceedd49258d42 from my transfer-pool branch? (This commit did not fix the problem on Linux.)

xlz:
Further testing shows a886bb0 introduces regression on Linux. Only one out of 60 ISO transfer actually completes and the data received is corrupted.

larshg:
Hi @xlz 

I tried the [commit](https://github.com/xlz/libfreenect2/commit/c6f6e57df3671285812e5ae3ac9ceedd49258d42) and it seems to work fine at my end.

Atleast I cannot provoke the iterator errors.

floe:
Just to clarify, this works on Windows & Linux now and can be merged?

xlz:
This can't be merged in the current form. There are two issues: use of C++11 feature std::mutex, and race condition of iterators of `idle_transfers_` in `TransferPool::deallocate()` and `TransferPool::onTransferComplete`.

larshg:
@floe I haven't tried on linux, as I don't have dual boot (yet).
I have tried to address the two things @xlz  pointed out. Using ```libfreenect2/threading.h``` and the mutex defined there and wrapped ```idle_transfers_``` iterator in a lock.

xlz:
Well, write access to `idle_transfers_` in `TransferPool::submit()` and `TransferPool::onTransferComplete()` is not locked yet. The problem here is `TransferPool::deallocate()` must only be called after all asynchronous transfer cancellation requests complete. If `deallocate()` is called before `onTransferComplete()`, callback function is applied on freed memory, and then results in crash, even if the iterator is protected by locks. I thought using only locks would solve this, but apparently not.

RyanGordon:
@xlz Would a semaphore then be more appropriate for handling that scenario?

hanyazou:
Pending_transfers_lock_.lock() and unlock() are needed in TransferPool::cancel(). It works well on my Mac. Without this, BAD ACCESS will occur when you push ESC key to quit the Protonect.

Process 14826 stopped
* thread #1: tid = 0x1289ea, 0x000000010001499e libfreenect2.dylib`libfreenect2::usb::TransferPool::cancel(this=0x00000001036066e8) + 142 at transfer_pool.cpp:120, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
    frame #0: 0x000000010001499e libfreenect2.dylib`libfreenect2::usb::TransferPool::cancel(this=0x00000001036066e8) + 142 at transfer_pool.cpp:120
   117  {
   118    for(TransferQueue::iterator it = pending_transfers_.begin(); it != pending_transfers_.end(); ++it)
   119    {
-> 120      int r = libusb_cancel_transfer(*it);
   121
   122      if(r != LIBUSB_SUCCESS)
   123      {
(lldb) bt
* thread #1: tid = 0x1289ea, 0x000000010001499e libfreenect2.dylib`libfreenect2::usb::TransferPool::cancel(this=0x00000001036066e8) + 142 at transfer_pool.cpp:120, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
  * frame #0: 0x000000010001499e libfreenect2.dylib`libfreenect2::usb::TransferPool::cancel(this=0x00000001036066e8) + 142 at transfer_pool.cpp:120
    frame #1: 0x00000001000235fb libfreenect2.dylib`libfreenect2::Freenect2DeviceImpl::stop(this=0x00000001036066c0) + 411 at libfreenect2.cpp:566
    frame #2: 0x0000000100007eca Protonect`main(argc=<unavailable>, argv=<unavailable>) + 2938 at Protonect.cpp:99
    frame #3: 0x00007fff8deb25fd libdyld.dylib`start + 1

larshg:
@hanyazou I have added the lock - can you see/test if it works?

hanyazou:
@larshg it works. thank you for you quick response.

floe:
In the current patch, `idle_transfers_` is only locked and unlocked exactly once (in `TransferPool::deallocate()`). Is this lock still required, then?

xlz:
@RyanGordon https://github.com/xlz/libfreenect2/commit/c6f6e57df3671285812e5ae3ac9ceedd49258d42 is my proposal to fix this bug. It refactors `TransferPool` using std::vector and waits for all transfers during cancellation (with a `stopped` flag). No locks are needed.

floe:
Superseded by lock-less variant in #275, closing.

PR #221, opened 2015-05-04
Improve RGB and depth stream parsers

xlz:
In the RGB stream parser, add an extra check using information the RGB packet footer revealed by new protocol analysis to detect JPEG EOI marker. Then use the marker's position to derive a correct JPEG length for some hardware JPEG decoders that can't handle garbage after the end of JPEG image.

In the depth stream parser, remove the scanning of a 8-byte magic footer which is very unreliable (the particular 8 bytes may appear in the middle) and inefficient. Based on [analysis of the isochronous transfer protocol](https://github.com/OpenKinect/libfreenect2/pull/75#issuecomment-98205817), assume packets have fixed size and the footer always resides at the end of a fixed sized packet. The changes made are minimized, and the execution order in the original parser is preserved, including the order of memcpy's and subsequence/sequence assembling order.

The code has been tested on Linux and Windows.

### DIFF
--- a/examples/protonect/include/libfreenect2/depth_packet_processor.h
+++ b/examples/protonect/include/libfreenect2/depth_packet_processor.h
@@ -40,6 +40,7 @@ namespace libfreenect2
 struct LIBFREENECT2_API DepthPacket
 {
   uint32_t sequence;
+  uint32_t timestamp;
   unsigned char *buffer;
   size_t buffer_length;
 };
--- a/examples/protonect/include/libfreenect2/frame_listener.hpp
+++ b/examples/protonect/include/libfreenect2/frame_listener.hpp
@@ -28,6 +28,7 @@
 #define FRAME_LISTENER_HPP_
 
 #include <cstddef>
+#include <stdint.h>
 #include <libfreenect2/config.h>
 
 namespace libfreenect2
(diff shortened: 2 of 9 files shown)


christiankerl:
Nice +1 i guess this subsumes the other packet parsing related PRS? However, timestamps are added but not passed through parsers/processors or did I miss something? Could you fix this otherwise?

xlz:
@christiankerl 
I have reorganized the commits so that there is a standalone commit to fully deal with timestamps and sequence numbers.

The RGB commit subsumes #72 and #73. The depth commit subsumes #75. The timestamp/sequence commit subsumes #71 and #148.

larshg:
This PR works fine at my end :+1: 

Only cosmetic is to remove the [old comment](https://github.com/OpenKinect/libfreenect2/pull/221/files#diff-f894f9a4db36f00a4e01f8e04e3c331cR102)?

hovren:
I did a quick check, and it seems to be working fine on my end as well (Linux). (y)

floe:
Also working fine for me (Ubuntu 14.04).

xlz:
Merge conflicts from #227 which changes cosmetic log strings. To fix later.

floe:
Sorry, didn't realize that - should I revert #227?

xlz:
No. Fixed conflicts.

wiedemeyer:
I just tested this pull request and it also solves the issue of restarting the device.

floe:
OK, sounds like we should integrate this now. I'll give it until today in the evening for any last-minute objections, then I'll merge it.

Issue #11, opened 2014-03-10
Remove boost thread dependency


christiankerl:
Threading should be done without boost. One option is to use pthreads with a minimal wrapper for Windows. Alternatively, include one of the following threading libs:
- http://tinythreadpp.bitsnbites.eu/
- http://tinycthread.bitsnbites.eu/

Issue #10, opened 2014-03-05
Define public API


christiankerl:
As we have the basic functionality now, the next step should be to define a public API. People could start working on nice applications using libfreenect2 and we can continue tinkering on the internals.

I guess most people would prefer a C API as in libfreenect for maximum portability? The functions should probably resemble the libfreenect API so switching is as seamless as possible? Are there any flaws in the libfreenect API, which should be avoided in  libfreenect2?

PR #226, opened 2015-05-07
add convenience method & sample code for registration

floe:
Thanks again @larshg and @xlz for your sample code. I've rewritten it to not use OpenCV, as we probably want to remove that dependency at some point (see also the PR by @christiankerl). Also, I have not yet implemented the depth buffer from @xlz's example, will do that soon.


xlz:
> is rounding the right way to go here - no subpixel precision?

RGB images are much oversampled than depth images. More sampling precision on RGB would have little improvement.

>      r_off = d_off*3;

3 is just hardcoded `bytes_per_pixel` value. VA-API and Tegra hardware jpeg decoder return in BGRA format. They would need 4 for `bytes_per_pixel`.

floe:
Good point about `bytes_per_pixel`, I'll fix that.

floe:
Any test reports from Windows or MacOS? I'm reluctant to merge this otherwise.

gaborpapp:
Thanks for your work on this. I tested on MacOS 10.9.5.  Registration seems to work, although I get some noise at the bottom of the image.

screen shot 2015-05-10 at 20 43 06

Also I get a crash on exit, which might not be related, but I don't get it with the latest 585d4c4 upstream/master.
Crash report:

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libusb-1.0.0.dylib              0x00000001091a0886 usbi_log_v + 70 (core.c:2290)
1   libusb-1.0.0.dylib              0x000000010919eb5f usbi_log + 127 (core.c:2378)
2   libusb-1.0.0.dylib              0x00000001091a9e76 darwin_cancel_transfer + 438 (darwin_usb.c:1746)
3   libusb-1.0.0.dylib              0x00000001091a3163 libusb_cancel_transfer + 99 (io.c:1500)
4   libfreenect2.dylib              0x0000000108f8128a libfreenect2::usb::TransferPool::cancel() + 138 (deque:299)
5   libfreenect2.dylib              0x0000000108f902cb libfreenect2::Freenect2DeviceImpl::stop() + 411 (libfreenect2.cpp:567)
6   Protonect                       0x0000000108f6eb99 main + 3033 (Protonect.cpp:106)
7   libdyld.dylib                   0x00007fff876a15fd start + 1

xlz:
The noise:
```
    uint8_t registered[depth->height*depth->width*3];
    registration.apply(rgb,depth,registered);
    cv::imshow("registered", cv::Mat(depth->height, depth->width, CV_8UC3, registered));
```
Need a memset here, or `registered = cv::Mat::zeros()`.

The crash is a race condition, discussed in #212.

gaborpapp:
Thanks! memset and the xlz@c6f6e57 commit mentioned in #212 indeed fixed both issues.

floe:
Thanks for testing - instead of memset, I've modified the inner loop to also set skipped pixels to zero, so we don't need to memset everything to zeroes which will get overwritten anyway.

gaborpapp:
Thanks. This also solves the issue.

xlz:
FTBFS with MSVC 2013 here:

    uint8_t registered[depth->height*depth->width*rgb->bytes_per_pixel];

Error C2057: expected constant expression

floe:
Ooops, OK (a bit surprising, but as you said, it's probably a better idea to move the buffer to the heap anyway). I'll fix that ASAP.

xlz:
Also, this line causes stack overflow with MSVC:

    libfreenect2::Registration registration(dev->getIrCameraParams(), dev->getColorCameraParams());

After replacing this and registered with heap allocation, it works correctly on Windows.

floe:
> Also, this line causes stack overflow with MSVC:

Now that is somewhat surprising, why should this line be an issue? IMHO it should be valid C++, can you post a stacktrace?

xlz:
The stack overflow happens upon entering main().

Probably because `libfreenect2::Registration` is quite large: 3392KB.

floe:
Ah, OK. I would have assumed the default stack is a lot bigger, but it's fixed now.

larshg:
I had to add ```#include <stdint.h> ``` to make it work for VS2013 Update 4 else ```uint8_t``` is unknown.

floe:
Also fixed, now using `unsigned char*` which is probably the most portable solution.

larshg:
Works great :+1:

examples/protonect/Protonect.cpp
```diff
@@ -87,6 +91,10 @@ int main(int argc, char *argv[])
     cv::imshow("ir", cv::Mat(ir->height, ir->width, CV_32FC1, ir->data) / 20000.0f);
     cv::imshow("depth", cv::Mat(depth->height, depth->width, CV_32FC1, depth->data) / 4500.0f);
 
+    if (!registered) registered = new unsigned char[depth->height*depth->width*rgb->bytes_per_pixel];
```
JuantAldea: Is there any reason to not to move this outside of the loop and get rid of the condition?

JuantAldea: Given that the depth frame size is already know to be 512x424 if I recall correctly, is there any reason to not to move the allocation out of the loop?

floe: Depending on the color frame decoder, rgb->bits_per_pixel is either 3 or 4, so I'd like to keep it consistent.

examples/protonect/Protonect.cpp
```diff
@@ -87,6 +91,10 @@ int main(int argc, char *argv[])
...
+    registration->apply(rgb,depth,registered);
+    cv::imshow("registered", cv::Mat(depth->height, depth->width, CV_8UC3, registered));
```
larshg: This line should also handle 3 or 4 bytes per pixel?

floe: Now that I think about it, it would probably be better to switch to 4 BPP anyway and use the extra channel to add depth data. That would make filtering a lot easier, and would probably also be faster than shifting around single bytes.

xlz: This line can be generalized when JPEG decoders of 4-byte output are added.

floe:
OK, since everything works as expected across all major platforms, I'll merge this tomorrow; additional features will be added via separate PRs. Thanks for contributing, everyone!

JuantAldea:
This are the results I am getting from the registration:

i) noticeable radial distortion in the closet. Shouldn't the depth data be undistorted first?

ii) pixel duplication, you can see my hand and the pen appear two times. Is this normal? Shouldn't registration be a bijection?

pixel_duplication_radial_distortion

 I'm quite newbie on this subject so I might be missing something.

floe:
@JuantAldea 
1) the depth data is already being undistorted, but the built-in parameters are probably just not as good as those you would get from an individual chessboard calibration. In any case, if you compare it to the raw depth data, I think the distortion is already a lot less.
2) this is a fundamental geometric limitation, from the POV of the color camera, the pixels on the right are directly behind the pen. This will be solved in a second PR by using a depth buffer.

JuantAldea:
@floe Thanks.

I'd say it looks like the depth isn't undistorted but the color is "distorted" so that it matches the depth.

I've superimposed the three images (registration, depth,  and IR) as layers on this [1] Gimp file. You can play a little with the transparency and the layer order.

[1] https://www.dropbox.com/s/t2nnf737jvt4b3y/IR_DEPTH_REGISTERED.xcf?dl=0

floe:
@JuantAldea good point, got it backwards myself. Nice example, thanks!

JuantAldea:
So is this working as intended? Shouldn't the depth be undistorted?

floe:
As far as we know, the parameters stored in the device are specific for this operation only, i.e. mapping color onto depth. Unless @sh0 knows about some more internals (such as the tables at the end of DepthCameraParamsResponse), this is all we can currently support without external calibration.

JuantAldea:
Got it, thanks.

floe:
Actually, maybe I was posting too soon: in theory, the parameters for standalone depth undistortion should be there, too. I'll re-read the old discussion in #41 tomorrow.

floe:
I was obviously posting this too late at night yesterday evening, of course it's possible to undistort the depth image separately - that's exactly what I added undistort_map for in the first place :-)

lakehanne:
I'm trying to do a face detection with the Protonect.cpp code. I need to apply a haarclassifier on the registered image. I tried with opencv haarcascade class 

>int key = cv::waitKey(1);
    protonect_shutdown = protonect_shutdown || (key > 0 && ((key & 0xFF) == 27)); // shutdown on escape

    listener.release(frames);
    //libfreenect2::this_thread::sleep_for(libfreenect2::chrono::milliseconds(100));

      //-- 1. Load the cascades
  if( !face_cascade.load( face_cascade_name ) ){ printf("--(!)Error loading\n"); return -1; };
  if( !eyes_cascade.load( eyes_cascade_name ) ){ printf("--(!)Error loading\n"); return -1; };

   //-- 2. Read the registration stream
  registration.open( -1 );
  if( registration.isOpened() )
  {
    for(;;)
    {
      registration >> frame;

      //-- 3. Apply the classifier to the frame
      if( !frame.empty() )
       { detectAndDisplay( frame ); }
      else
       { printf(" --(!) No captured frame -- Break!"); break; }

      int c = waitKey(10);
      if( (char)c == 'c' ) { break; }

    }
  }
  }

  // TODO: restarting ir stream doesn't work!
  // TODO: bad things will happen, if frame listeners are freed before dev->stop() :(
  dev->stop();
  dev->close();

  delete registration;

  return 0;
}

/**
 * @function detectAndDisplay
 */
void detectAndDisplay( Mat frame )
{
   std::vector<Rect> faces;
   Mat frame_gray;

   cvtColor( frame, frame_gray, COLOR_BGR2GRAY );
   equalizeHist( frame_gray, frame_gray );
   //-- Detect faces
   face_cascade.detectMultiScale( frame_gray, faces, 1.1, 2, 0|CV_HAAR_SCALE_IMAGE, Size(30, 30) );

   for( size_t i = 0; i < faces.size(); i++ )
    {
      Point center( faces[i].x + faces[i].width/2, faces[i].y + faces[i].height/2 );
      ellipse( frame, center, Size( faces[i].width/2, faces[i].height/2), 0, 0, 360, Scalar( 255, 0, 255 ), 2, 8, 0 );

      Mat faceROI = frame_gray( faces[i] );
      std::vector<Rect> eyes;

      //-- In each face, detect eyes
      eyes_cascade.detectMultiScale( faceROI, eyes, 1.1, 2, 0 |CV_HAAR_SCALE_IMAGE, Size(30, 30) );

      for( size_t j = 0; j < eyes.size(); j++ )
       {
         Point eye_center( faces[i].x + eyes[j].x + eyes[j].width/2, faces[i].y + eyes[j].y + eyes[j].height/2 );
         int radius = cvRound( (eyes[j].width + eyes[j].height)*0.25 );
         circle( frame, eye_center, radius, Scalar( 255, 0, 0 ), 3, 8, 0 );
       }
    }
   //-- Show what you got
   imshow( window_name, frame );

but I kept getting a member error `error: request for member ‘open’ in ‘registration’, which is of pointer type ‘libfreenect2::Registration*’ (maybe you meant to use ‘->’ ?)
   registration.open( -1 );
`
What is the correct way to access the registered/depth/rgb/undistorted frames?

floe:
@lakehanne see here: https://github.com/OpenKinect/libfreenect2/blob/master/examples/protonect/Protonect.cpp#L149-L152

lakehanne:
Thanks!

lakehanne:
Maybe I could use your help but the Protonect object keeps crashing during runtime with 

>pure virtual method called
terminate called without an active exception
Aborted (core dumped).

 Here's the code [gist](https://gist.github.com/lakehanne/b601f96c07d1b02ede2c). Is there something I am missing? I used gdb examine the core file and here's what I found: 

>     #0  0x00007ffff63a9cc9 in __GI_raise (sig=sig@entry=6)
        at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
        #1  0x00007ffff63ad0d8 in __GI_abort () at abort.c:89
        #2  0x00007ffff69ae535 in __gnu_cxx::__verbose_terminate_handler()    ()
        from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
        #3  0x00007ffff69ac6d6 in ?? () from /usr/lib/x86_64-linux-    gnu/libstdc++.so.6
        #4  0x00007ffff69ac703 in std::terminate() ()
         from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
        #5  0x00007ffff69ad1bf in __cxa_pure_virtual ()
          from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
        #6  0x00007ffff7a0bbbd in    libfreenect2::OpenGLDepthPacketProcessor::process (this=0x71c490, packet=...)
           at    /home/lex/Desktop/libfreenect2/examples/protonect/src/opengl_depth_packet_ processor.cpp:948
        #7  0x00007ffff79fa61d in execute (this=0x13be3a0)
          at    /home/lex/Desktop/libfreenect2/examples/protonect/include/libfreenect2/asy nc_packet_processor.h:105
       #8      libfreenect2::AsyncPacketProcessor<libfreenect2::DepthPacket>::static_execute (data=0x13be3a0)
        at      /home/lex/Desktop/libfreenect2/examples/protonect/include/libfreenect2/asy nc_packet_processor.h:91
        #9  0x00007ffff7a0904f in tthread::thread::wrapper_function  (aArg=0x13b1e00)
          at    /home/lex/Desktop/libfreenect2/examples/protonect/src/tinythread/tinythre-       --Type <return> to continue, or q <return> to quit---
ad.cpp:169
#10 0x00007ffff5ceb182 in start_thread (arg=0x7fffd5d75700)
    at pthread_create.c:312
#11 0x00007ffff646d47d in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111

PR #236, opened 2015-05-20
Extension of Protonect to allow selection of pipeline and device

wiedemeyer:
I just added some code to `Protonect` to allow the selection of the pipeline and the device via parameters. This makes it easier to check whether certain pipelines work on a system or not. Without any parameters it behaves like before.

Usage: `./Protonect [pipeline] [serial]` (the order don't matter)
Parameters for the pipelines are: `cpu`, `gl`, `cl`
Every parameter that just contains numbers is used as serial number: `012345678901` for example.

### DIFF
--- a/examples/protonect/Protonect.cpp
+++ b/examples/protonect/Protonect.cpp
@@ -34,6 +34,7 @@
 #include <libfreenect2/frame_listener_impl.h>
 #include <libfreenect2/threading.h>
 #include <libfreenect2/registration.h>
+#include <libfreenect2/packet_pipeline.h>
 
 bool protonect_shutdown = false;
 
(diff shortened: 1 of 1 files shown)


wiedemeyer:
I had it like that before, but changed it, because the reading of the parameter looked a bit overloaded with all the `#ifdef #else #endif` things. So I decided to split it up in one part that does just the reading of the parameter and one that applies them. But I can change it if desired.
And I will fix the typo. ;-)

xlz:
You still have ifdefs in the switch. The translation will mean 2x the code when we have 3 rgb processors, 4 depth processors, 3 registration processors, with their combinations...

I agree mixing ifdefs in code is an eyesore. Perhaps we need a `libfreenect2::createXXXPacketPipeline()` and move `LIBFREENECT2_WITH_XXX_SUPPORT` into that function.

wiedemeyer:
I updated the pull request.

xlz:
Tests OK.

floe:
I'll merge this tomorrow, as it will hopefully help a lot in addressing issues where specific depth processors don't work.

PR #238, opened 2015-05-21
commented out unused variable from cpu_depth_packet_processor.cpp

gaborpapp:
Can this be removed or something more complex is missing?

### DIFF
--- a/examples/protonect/src/cpu_depth_packet_processor.cpp
+++ b/examples/protonect/src/cpu_depth_packet_processor.cpp
@@ -580,9 +580,7 @@ public:
         {
           if(max_edge_test_ok)
           {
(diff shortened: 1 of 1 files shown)


floe:
Since this is a leftover of the conversion from shader to CPU code, I'd prefer to just comment it out.

xlz:
Just leave it and compiler will optimize it out for you.

gaborpapp:
Sure, but it gives a warning every time I compile the code, which is not very elegant.

gaborpapp:
Commented the variable out instead as you prefer.

floe:
Yeah, I think that's the best solution. Merged.

PR #240, opened 2015-05-22
add a brief (linux-centric) FAQ section

floe:


RyanGordon:
Some of these instructions are Linux specific, it would be cool if we mentioned that the FAQ is specific to Linux or the individual instructions where needed

floe:
I think in most places, the Linux-specific parts are actually marked as such. Could you perhaps provide Windows-specific additions for the FAQ entries?

RyanGordon:
Sorry, you're right! Not sure how I missed that

PR #241, opened 2015-05-22
Renamed contrib folder to rules

larshg:
Windows can't handle a folder with same name as a file with no extensions.

due to #239 @floe


floe:
Ah crud... thanks for catching that.

Issue #223, opened 2015-05-05
Registration


giacomodabisias:
Hi,
I am trying to apply the new registration you did with the new parameters taken directly from the kinect2.

I created the Registration object and took the parameters from the device with the 
getIrCameraParams()
getColorCameraParams()

functions.
Now I have to use the apply function of the class, but I am wondering what size the two images should have. Should I resize the rgb to the depth size or can I use the function without resizing the images?

Thanks

xlz:
#190 has demos showing how to use the registration class. Code is here https://github.com/larshg/libfreenect2/tree/registration .

floe:
Thanks @xlz - I will hopefully get around to integrating these examples into the master branch soon (a bit swamped with work at the moment :-).

giacomodabisias:
I tried the registration, but I assume there must be something wrong in the parameters probably (?)
Here is the cloud created with the registration:
![bad2](https://cloud.githubusercontent.com/assets/5880012/7491553/a1686fa8-f3ea-11e4-847a-494d11e696c8.png)

while the one made with my own calibration looks like this:
![good](https://cloud.githubusercontent.com/assets/5880012/7491560/ac3125ec-f3ea-11e4-9660-df7c08e73f7b.png)

I could create a small video to see the difference if needed.

floe:
Can you describe in a bit more detail how they differ? They look quite similar to me... but in general, a proper chessboard calibration will usually have higher quality than the factory calibration, which is IMHO optimized for speed.

giacomodabisias:
I will prepare a short video, so you can see better. If you enlarge the two
images and look at the hand you will see a lot of noise

2015-05-06 14:09 GMT+02:00 Florian Echtler <notifications@github.com>:

> Can you describe in a bit more detail how they differ? They look quite
> similar to me... but in general, a proper chessboard calibration will
> usually have higher quality than the factory calibration, which is IMHO
> optimized for speed.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/OpenKinect/libfreenect2/issues/223#issuecomment-99426634>
> .
>



-- 
G. Dabisias, PhD Student
PERCRO, (Laboratory of Perceptual Robotics)
Scuola Superiore Sant'Anna
via Luigi Alamanni 13D, San Giuliano Terme 56010 (PI), Italy
mob.: +39 3480839095

floe:
[...]

PR #253, opened 2015-06-02
Added filtering of shadowed color regions to registration

wiedemeyer:
Hi,

I tested the registration and it works fine, nice work! But it ran a bit slow on my system, so I decided to improve it. The original version took ~5.1 ms on my system. With this changes it went down to ~1 ms.

I achieved this by replacing the multi dimensional array with a one dimensional one, which is structured similar to the image data. `map[i]` is the value for `depth[i]`, etc. So you only need to walk through the array with a pointer and you are done.

The new maps also include some static multiplications and additions which where computed at runtime before. There is also a new map for y image offsets, to reduce computations while runtime even more.

The rounding of the x coordinate is replaced by `const int cx = rx + 0.5f;`, which should be the same for positive numbers. But all valid x coordinates should be greater equals 0 anyway. To make sure that x is positive, a check for `rx > -0.5f` is added to the if statement below.

The last thing I changed is the API of the `apply` method. I changed the type for the registered image to libfreenect2::Frame, so that it is possible to check for the correct size.

### DIFF
--- a/examples/protonect/src/registration.cpp
+++ b/examples/protonect/src/registration.cpp
@@ -100,48 +100,107 @@ void Registration::apply(const Frame *rgb, const Frame *depth, Frame *undistorte
   const int *map_dist = distort_map;
   const float *map_x = depth_to_color_map_x;
   const int *map_yi = depth_to_color_map_yi;
+
   const int size_depth = 512 * 424;
-  const int size_color = 1920 * 1080 * 3;
+  const int size_color = 1920 * 1080;
   const float color_cx = color.cx + 0.5f; // 0.5f added for later rounding
 
+  // size of filter map with a border of filter_height_half on top and bottom so that no check for borders is needed.
+  // since the color image is wide angle no border to the sides is needed.
+  const int size_filter_map = size_color + 1920 * filter_height_half * 2;
+  // offset to the important data
+  const int offset_filter_map = 1920 * filter_height_half;
+
+  // map for storing the min z values used for each color pixel
+  float *filter_map = new float[size_filter_map];
+  // pointer to the beginning of the important data
+  float *p_filter_map = filter_map + offset_filter_map;
+
+  // map for storing the color offest for each depth pixel
+  int *depth_to_c_off = new int[size_depth];
+  int *map_c_off = depth_to_c_off;
+
+  // initializing the depth_map with values outside of the Kinect2 range
+  for(float *it = filter_map, *end = filter_map + size_filter_map; it != end; ++it){
+    *it = 65536.0f;
+  }
+
   // iterating over all pixels from undistorted depth and registered color image
-  // the three maps have the same structure as the images, so their pointers are increased each iteration as well
-  for (int i = 0; i < size_depth; ++i, ++registered_data, ++undistorted_data, ++map_dist, ++map_x, ++map_yi) {
+  // the four maps have the same structure as the images, so their pointers are increased each iteration as well
+  for(int i = 0; i < size_depth; ++i, ++undistorted_data, ++map_dist, ++map_x, ++map_yi, ++map_c_off){
     // getting index of distorted depth pixel
     const int index = *map_dist;
 
     // check if distorted depth pixel is outside of the depth image
     if(index < 0){
+      *map_c_off = -1;
       *undistorted_data = 0;
-      *registered_data = 0;
-      *++registered_data = 0;
-      *++registered_data = 0;
       continue;
     }
 
     // getting depth value for current pixel
-    const float z_raw = depth_data[index];
-    *undistorted_data = z_raw;
+    const float z = depth_data[index];
+    *undistorted_data = z;
 
     // checking for invalid depth value
-    if (z_raw <= 0.0f) {
+    if(z <= 0.0f){
+      *map_c_off = -1;
+      continue;
+    }
+
+    // calculating x offset for rgb image based on depth value
+    const float rx = (*map_x + (color.shift_m / z)) * color.fx + color_cx;
+    const int cx = rx; // same as round for positive numbers (0.5f was already added to color_cx)
+    // getting y offset for depth image
+    const int cy = *map_yi;
+    // combining offsets
+    const int c_off = cx + cy * 1920;
+
+    // check if c_off is outside of rgb image
+    // checking rx/cx is not needed because the color image is much wider then the depth image
+    if(c_off < 0 || c_off >= size_color){
+      *map_c_off = -1;
+      continue;
+    }
+
+    // setting a window around the filter map pixel corresponding to the color pixel with the current z value
+    int yi = (cy - filter_height_half) * 1920 + cx - filter_width_half; // index of first pixel to set
+    for(int r = -filter_height_half; r <= filter_height_half; ++r, yi += 1920) // index increased by a full row each iteration
+    {
+      float *it = p_filter_map + yi;
+      for(int c = -filter_width_half; c <= filter_width_half; ++c, ++it)
+      {
+        // only set if the current z is smaller
+        if(z < *it)
+          *it = z;
+      }
+    }
+
+    // saving the offset for later
+    *map_c_off = c_off;
+  }
+
+  // reseting the pointers to the beginning
+  undistorted_data = (float*)undistorted->data;
+  map_c_off = depth_to_c_off;
+
+  // run through all registered color pixels and set them based on filter results
+  for(int i = 0; i < size_depth; ++i, ++registered_data, ++map_c_off, ++undistorted_data){
+    const int c_off = *map_c_off;
+
+    // check if offset is out of image
+    if(c_off < 0){
       *registered_data = 0;
       *++registered_data = 0;
       *++registered_data = 0;
       continue;
     }
 
-    // calculating x offset for rgb image based on depth value
-    const float rx = (*map_x + (color.shift_m / z_raw)) * color.fx + color_cx;
-    const int cx = rx; // same as round for positive numbers (0.5f was already added to color_cx)
-    // getting y offset for depth image
-    const int cy = *map_yi;
-    // combining offsets
-    const int c_off = cx * 3 + cy;
+    const float min_z = p_filter_map[c_off];
+    const float z = *undistorted_data;
 
-    // check if c_off is outside of rgb image
-    // checking rx/cx is not needed because the color image is much wider then the depth image
-    if (c_off < 0 || c_off >= size_color) {
+    // check for allowed depth noise
+    if((z - min_z) / z > filter_tolerance) {
       *registered_data = 0;
       *++registered_data = 0;
       *++registered_data = 0;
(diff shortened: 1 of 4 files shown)


floe:
Awesome, I was planning for something similar, but I wouldn't have assumed that it's actually that fast. I'll merge this soon-ish.

examples/protonect/src/registration.cpp
```diff
@@ -72,60 +72,62 @@ void Registration::depth_to_color(float mx, float my, float& rx, float& ry)
     (mx * mx * color.my_x2y0) + (my * my * color.my_x0y2) + (mx * my * color.my_x1y1) +
     (mx * color.my_x1y0) + (my * color.my_x0y1) + (color.my_x0y0);
 
-  rx = wx / (color.fx * color_q);
-  ry = wy / (color.fx * color_q);
+  rx = (wx / (color.fx * color_q)) - (color.shift_m / color.shift_d);
+  ry = (wy / (color.fx * color_q)) * color.fy + color.cy;
```
floe: Should this be `color.fy` in the second line? Then it could also be simplified to `ry = (wy / color_q) + color.cy;`

wiedemeyer: I think it should be `fy`, I didn't checked the previous code for errors. But from the senser provided intrinsics `fy` and `fx` are always the same. But if someone uses other intrinsics this could lead to wrong values, I will change it.

wiedemeyer:
I was impressed too. After restructuring the maps and using pointers it went down to 2.3 ms. With the static computations added to the maps it went down to 1.9 ms. With the extra map for `cy` and `floorf(rx + 0.5f)` instead of `round(rx)`, it was 1.3 ms. And finally after replacing `floorf(rx + 0.5f)` by `(int)(rx + 0.5f)` it was only 1 ms.
I think now it is so fast, that there is no need for an OpenCL/GL/CUDA implementation (at least on fast x86 machines). I am afraid, that copying the color and depth image to the GPU and reading the result might consume a lot of this 1 ms, so there is less headroom to improve.

xlz:
Does this include depth filtering? I think that's the hard part, involving a depth map of 1920x1080 (or half the size).

floe:
@xlz you mean depth buffer? Not yet. I have the undistortion working now, unfortunately in a separate branch. Currently checking what's the best way to integrate this PR with my code.

wiedemeyer:
Well, this is now only a color -> depth registration, so there won't be a high resolution depth map. I though about a depth -> color registration using the inverse of this registration, but didn't found a way to do this, while `cy` is static for each `x,y`, the depth value is involved to in the middle of the calculation to find the corresponding `cx` coordinate.
For the filtering I thought, one could create an array of the size of the color image and store for each pixel if it has been used for the registered color image (and for which registered color pixel it has been used). If two registered color pixels are using the same color image pixel, just select the one that belongs to the depth pixel with the lower distance. But the problem is that the color image is much bigger, so that the background and foreground pixels could just map to neighboring color pixels and then you won't find out, that there are background pixel. So for the filtering we need something more fancy.

floe:
@xlz has a branch where a depth buffer has already been implemented, see his comment in #223.

xlz:
This one does depth filter: https://github.com/xlz/libfreenect2/commit/a0c0b6e5402ad363dcd5bbe58ebab8d093ea2f80

What I meant for depth map is an intermediate map to filter duplicate pixels, just as you described. The fancy something is erosion which extracts minimum of neighboring pixels. Small erosion kernel size will produce aliasing in the edge. There was unavoidable aliasing with erosion on a half-size depth filter so I chose full-size.

In depth to color registration, it is upsampling the depth image so higher depth resolution wouldn't mean higher depth quality. In color to depth registration, even naive downsampling of RGB image will still retain very RGB quality in the pointcloud.

wiedemeyer:
Nice. I though of something like this, setting a whole neighborhood instead of one pixel. But this filtering is pretty expensive, especially the erode.

Yeah, you can't add information when upsampling, but you will get more detailed texture in the point cloud. It always depends on what you want.

wiedemeyer:
I updated the pull request. It now includes undistortion of the depth and registered color image. It is still arround 0.9 to 1.0 ms. on my system. There was also a double conversion in one of the if statements, which I also replaced.
Maybe we can also build in the filtering. Shouldn't be to hard to do this without opencv.

floe:
Great! Just one small additional request: could you as some comments to your code? There's a fair amount of pointer arithmetic going on, and it's rather easy to lose track...

wiedemeyer:
Added some comments, hope it is easier to keep track now.

wiedemeyer:
I just implemented the filtering from @xlz into the registration, but without any OpenCV. I tried to optimize the filtering as well. The hole registration with filtering takes now ~5.5 ms. on my system.

@xlz: could you check if the filtering is fine like this and if it works similar to your implementation?

xlz:
Good work. Unfortunately I won't be able to test it for a while due to traveling etc.

I roughly looked at the new code. Most looks good. One issue: don't assume 3 bytes per pixel. Hardware decoders (Tegra, Vaapi) will return RGBA format.

IIUC this is still a 5x3 filter on a 1920x1080 image. If you haven't done so, also verify if there is any aliasing  with maximum depth range.

HenningJ:
It does not always do what it's supposed to:
registration
Seems to work for my "shadow", but the shadow of the cup in front of me is not completely black. Is that because it's quite close to the camera?
And what's going on with the bottle? Is that because it's mostly transparent? 

But otherwise it looks pretty good

More general question: the (small) shadow on the left side (that's in the original depth image) is because of the distance between IR-emitter and IR-sensor? and the (larger) shadow on the right side (in the registered image) is because of the distance between IR-sensor and RGB-sensor? Is that correct?

wiedemeyer:
@xlz ok, didn't know that there are rgba color images. Is the byte order `RGBA`?
Yes, it's a 5 x 3 filter. Should be the same as or at least similar to your implementation. Instead of eroding later, it's done in one step.
How do I check for aliasing? Could an adaptive filter that is bigger for near pixels and smaller for far ones solve those issues?
I also want to make the filtering optional, so that it can be disabled if not needed.

@HenningJ the filter only works if there are depth values. At the borders of objects, like the cup, you don't get depth values, so there are no foreground depth pixel which use the color pixel from the border of the cup. Only the depth pixel from the background map to those border pixels, therefore they get used for those background depth pixels.
The same is true for the bottle. You can see that there are no values for the middle area (possibly to close to the camera), so the color pixels are used for the background pixels.
The shadow on the left is probably from the distance between the emitter and the sensor.

xlz:
RGBA yes. 

If you see nothing strange then no aliasing. Do not need fancy filtering here.

Optional sure.

floe:
What I did in my (unmerged) registration code is to simply treat all color image data as RGBA (i.e. registered color images are output as RGBA, and input images are assumed to be RGBA, too). Only difference for RGB images is that the source pointer is advanced by 3 instead of 4 bytes. IMHO, this has two advantages:
 * we can always use uint32_t operations, which are generally not slower, and often faster, than moving 3 separate bytes around
 * there is an additional data channel available which could be used for downsampled depth data.

However, I would not implement this within this current PR. @wiedemeyer if possible, I'd like to split this into several PRs:
 1. the original one (this one here, #253) doing just the speedup
 1. a separate one for the depth buffer
 1. and possible a third for parameters, tweaks etc.

floe:
I tried to cherry-pick the commits except the last one (97976ea), but apparently Github isn't smart enough to fast-forward the PR on its own. Can you try and see what happens if you merge master into this PR?

wiedemeyer:
I rebased it to the current master. But I saw you already pulled the first commits.

HenningJ:
@wiedemeyer thanks for the explanation.

wiedemeyer:
Just to make sure, the color image from the CPU JPEG endcoding has the `BGR` format, but VAAPI and Tegra have `RGBA`, right?
I just added the changes, so that the code can handle 3 and 4 byte color images. It also improved performance a bit (5.0 instead of 5.5 ms.).

wiedemeyer:
Yes, of course. I saw it right after I pushed.

HenningJ:
enough pestering, gotta get back to work ;-)

wiedemeyer:
If the input image is 3 byte the alpha channel of the output image will be set to the blue channel value of the next pixel.
First I though a 3 byte output image would be ok too, then the next run would overwrite the previous 4th byte anyway, but this will always write one byte after the end of the buffer, which might not be so good.

HenningJ:
> First I though a 3 byte output image would be ok too, then the next run would overwrite the previous 4th byte anyway

Yeah, that's what I meant.

> but this will always write one byte after the end of the buffer, which might not be so good.

This will happen now anyways, won't it? You're still writing a four byte (unsigned) int when there are only 3 bytes left in the 3 byte output image. You're just writing 0 to the additional byte, instead of some other number. If there is anything at that byte, it's gonna be bad either way. So I don't see any scenario where the additional &-operation actually helps.

wiedemeyer:
Ok, the `&` could be removed, if everyone ignores that byte anyway.
The current implementation only allows 4 byte output images. So that writing over the buffer is not an issue.
Maybe it would be nicer if all color images in libfreenect2 have the same format either 3 or 4 byte, then no special cases need to be handled. Especially users can be sure that the image format is always the same, independent from the JPEG processor that is being used.

HenningJ:
> Ok, the & could be removed, if everyone ignores that byte anyway.

Well...if anybody needs that byte to be 0, then of course your implementation is correct, it's just the comment thats wrong. ;-)

floe:
I think that we should simply change `turbo_jpeg_rgb_packet_processor.cpp`, lines 76 and 118, and replace TJPF_BGR by TJPF_RGBX. That would reduce the number of special cases quite a bit.

wiedemeyer:
I tried it out, and it makes things much more beautiful.
But if `TJPF_RGBX` is used, the images need to be converted to `BGRX` or `BGR` before then can be displayed with OpenCV in `Protonect`.
So it might be better to use `TJPF_BGRX` or even `TJPF_BGRA` (if we care about the values in the alpha channel). @xlz could VAAPI and Tegra also output `BGRA` instead of `RGBA`?

wiedemeyer:
should I push the changes to this pull request so you can try them out?

floe:
Yes, please. Regarding BGRA vs RGBA: we're trying to get rid of OpenCV anyway, the viewer from PR #261 is based on OpenGL and can use RGBA. But for the moment, TJPF_BGRX would probably be the best choice.

wiedemeyer:
I updated the PR.
I would vote for sticking with the BGRA format. But mainly because I am used to it. I did not find any information on which format is better. Only http://en.wikipedia.org/wiki/RGBA_color_space mentioned, that colors are commonly stored as ARGB in a 32 bit word, which results in BGRA on little endian.

floe:
OK, from what I've seen, any OpenGL implementation which can deal with libfreenect2 should also be able to handle BGRA just fine.

xlz:
VA-API can output both. Tegra can only output RGBA.

floe:
For the registration, the color order probably doesn't even matter now that we simply treat all individual pixels as `uint32_t`. @wiedemeyer Did you notice any speed issues with the color image decoder? After all, it needs to push 33% more data around now for each frame...

wiedemeyer:
Yeah, the order does not matter for the registration. The registration with filter takes ~5.7 ms. and 1.0 ms. without. So no real changes.
I didn't looked at the jpeg decoder, but I guess that moving 33% more bytes is just a fraction of the time needed to decode the jpeg data.

wiedemeyer:
Just let both versions run of the jpeg decoder run and both are around 14 ms.

floe:
Excellent, good to know. Merged.

PR #257, opened 2015-06-04
Streamlined the JPEG and added environment to work on linux/mac too.

larshg:
Added depends/libjpeg_turbo as search path


floe:
Looks good on my Ubuntu install, merged.

PR #260, opened 2015-06-05
Missing include and lib for default path on windows.

larshg:
Missing /include and /lib for depends folder.


gaborpapp:
Also works on OSX and finds libturbo_jpeg in the depends folder. thanks.

PR #262, opened 2015-06-06
build static library libfreenect2.a as well

hanyazou:
This mod allows to build shared and static library both. I've tested on my Mac, Xcode 6.2.


xlz:
@hanyazou 
I'm curious. What is the use case for building the static version? I think the final executable can't avoid being dynamic because many things are opened from dlopen().

hanyazou:
I wanted to build OpenNI2 driver "Freenect2Driver.dylib" binary distribution which did not require libfreenect2.dylib so that users could easily install the driver into somewhere OpenNI2 driver directory. I was afraid that multiple .dylib and library path issue is too difficult to handle by users. I however think I can solve this issue by setting RPATH properly.

PR #263, opened 2015-06-06
build static library libfreenect2.a as well

hanyazou:
This mod allows to build shared and static library both. I've tested on my Mac, Xcode 6.2.


xlz:
Does `cmake -DBUILD_SHARED_LIBS:BOOL=OFF` work for you?

hanyazou:
@xlz No that doesn't because of hard coded 'SHARED' in cmake files's "ADD_LIBRARY(freenect2 SHARED ${SOURCES})". You can choose 1) delete 'SHAED' in the cmake file or 2) build static and shared library like my mod. I think it definitely be convenient if it provide both simultaneously. I don't recommend you to build twice to prepare a distribution binary in the future.

xlz:
Well, I for one don't need the static version, then how can I disable it?

hanyazou:
So, shall we remove the 'SHARED' ? It's ok for me.

xlz:
I agree.

floe:
As it turns out, you can create both static and shared libraries without having to re-build all sources. Proposed replacement for this PR in #276.

PR #267, opened 2015-06-07
Better solution for dependency installation

floe:
On Ubuntu 14.04 and beyond, glfw3 can be installed from deb packages, so we should do that instead of relying on a self-built version. Also, install all other dependencies from the package manager & fixes directly from the script.


floe:
@xlz @gaborpapp @goldhoorn @larshg @wiedemeyer comments welcome.
TODO: install mac dependencies via brew directly from script?

xlz:
First, I'd rather not automate things that involve sudo. This should be loosely described in the documentation.

libglfw3-dev is available in Ubuntu vivid and later. For Ubuntu 14.04 trusty I recommend manually download libglfw3 and libglfw3-dev from official website (packages.ubuntu.com, but the user should be able to choose) and let the user `sudo dpkg -i`. 

For libturbojpeg.so, I guess there should a workaround that does not involves sudo that modifies things in `/usr`.

floe:
Excellent, thanks for the turbojpeg fix. Included. Also a good point about using sudo in scripts, I had some reservations myself. Alternative solution commited.

xlz:
OK there have been lots of commits. If you intend for more reviews please merge some of them into logical units.

floe:
Hmm, I'll try a rebase. Since there have been no line comments, this should still work.

floe:
Ok, rebasing didn't work because it's already public. I'll reopen a new PR.

xlz:
Instructions for installing glfw3 via brew:

```
brew tap homebrew/versions
brew install glfw3
```

It is installed in `/usr/local/opt/glfw3` with `lib/` `include/` and `lib/pkgconfig/glfw3.pc`. (Change FindGLFW3.cmake accordingly)

gaborpapp:
Using ports glfw3 gets installed in `/opt/local`:

```
port install glfw
```

xlz:
Support both then. Discussion continues in #271.

PR #270, opened 2015-06-07
Added depends search path.

larshg:
Removed old paths and text.

Added condition if debug is not found to set debug as the release library. Due to the "used" binary provided here: https://www.dropbox.com/s/madoye1ayaoajet/libusbx-winiso.zip


xlz:
Self-contained changes. Please merge so the rest of cmake clean up can continue.

floe:
Seems to be Windows-only, so I'll take your word for it :-)

PR #271, opened 2015-06-07
More modular solution for dependency installation

floe:
@xlz new PR with squashed/merged commits.


floe:
BTW, installing the packages from .deb files will also work on Vivid and later releases and will have the same effect as installing them via apt-get.

xlz:
Ubuntu 14.04 is a special case for long-term support. Debian/Ubuntu vivid+ all have included libglfw3-dev, and the user would want to install it via proper package management (provides updating).

On Mac OSX, install glfw3 with
```
brew tap homebrew/versions
brew install glfw3
```
And add `/usr/local/opt/glfw3/lib/pkgconfig` to PKG_CONFIG_PATH.

floe:
That's what I meant: even if you install the packages using `dpkg -i` on Vivid++, they will still be handled like all other packages and will also be updated.

Regarding the Mac stuff, as I don't have a machine to test it on, I'd prefer PRs from someone who can actually test them before. Been burned once already by the big cmake mess :-)

xlz:
Already then, looks good to me.

PR #275, opened 2015-06-08
Fix transfer pool thread safety

xlz:
Avoid unsafe access during transfer resubmission by refactoring TransferPool using std::vector. It seems there is no need to maintain a queue for the case when a transfer may temporarily fail and recover. If one transfer fails, it is likely all will fail permanently.

Each transfer is independent during runtime and no lock is needed.

Wait for all transfers during cancellation, when the only lock in the TransferPool protects the counter of stopped transfers. Though coarse-grained, this lock will not have effect on runtime performance.

This has been tested on Linux, Windows, and Mac. No crashes are observed, and shutdown time is reduced.

This PR proposes to subsume #212.

### DIFF
--- a/examples/protonect/include/libfreenect2/usb/transfer_pool.h
+++ b/examples/protonect/include/libfreenect2/usb/transfer_pool.h
@@ -27,10 +27,11 @@
 #ifndef TRANSFER_POOL_H_
 #define TRANSFER_POOL_H_
 
-#include <deque>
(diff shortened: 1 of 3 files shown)


gaborpapp:
Tested. Cleaner and faster shutdown. Seems to work well.

HenningJ:
Tested on Windows. Works great.

PR #276, opened 2015-06-08
create static and shared library from same source build

floe:


floe:
Note: works on Ubuntu 14.04, not yet tested on MacOS X and Windows.

gaborpapp:
Tested on OSX. The static and dynamic libs are built.

HenningJ:
Just tested in on Windows (with Visual Studio 2008). 

1. You renamed resources.inc everywhere it is used, but did not rename the file itself.
2. For me, both the shared and the static lib are built to the exact same place, with the same name. So one overwrites the other and you don't really know which of the libs you just built. Can't you build them to different folders or with different names?

HenningJ:
Oh, I just noticed the resources.inc.h isn't checked in, but gets generated. So, forget my first point. Sorry about that. :-)

xlz:
Under MSVC freenect2.lib is a linking wrapper for freenect2.dll. I suggest this change:

```diff
-set_target_properties(freenect2static PROPERTIES OUTPUT_NAME freenect2)
+IF(MSVC)
+  set_target_properties(freenect2static PROPERTIES OUTPUT_NAME freenect2static)
+ELSE()
+  set_target_properties(freenect2static PROPERTIES OUTPUT_NAME freenect2)
+ENDIF()
```
I think MinGW will output libfreenect2.a for static library so it should be enough to single out MSVC.

Tentative testing based on this change shows correct generation of libraries: freenect2.dll, freenect2.lib, freenect2static.lib.

HenningJ:
Yes, that sounds good :)

floe:
Fixed, with a slightly smaller changeset. Please re-test on Windows, then I'll merge.

HenningJ:
Yes, works on Windows/Visual Studio

PR #277, opened 2015-06-08
Add libusb as a path_suffixes - as libusb doesn't have a include folder.

larshg:


HenningJ:
Isn't that handled by the `"${DEPENDS_DIR}/libusb/libusb"` line in the `PATHS` above?

larshg:
Well, I haven't placed libusb in the depends folder, but else you are right. I have separate libraries folder, where I keep all the libraries I use for various projects :)

HenningJ:
I see. But then it would make sense to remove the second `libusb`from the line I quoted, since that would now be handled by the one in `PATH_SUFFIXES`, right? Not really necessary, but it would look nicer :-)

larshg:
Yeah, thats true - I'll correct it.

xlz:
This happened when I was fiddling the final patches and missed the `libusb` PATH_SUFFIXES. When testing Windows again it reported error, so I added an incorrect "typo" fix.

Issue #281, opened 2015-06-09
Why patch libusb?


tikiking1:
Simply put, why is libusb being patched? On Arch Linux, with libusb 1.0.19, Protonect encounters bug #182 after the original build tree (in `/tmp`) is deleted. On the same Arch Linux, requiring patched libraries is generally frowned upon.

gaborpapp:
according to readme.depends.txt it increases MAX_ISO_BUFFER_LENGTH, which is required for kinect2 to work as far as I know:

kernel 3.10+
libusbx 1.0.17 
  with superspeed patch by Joshua Blake <joshblake@gmail.com> see https://github.com/JoshBlake/libusbx/tree/superspeed
  with MAX_ISO_BUFFER_LENGTH increased to 49152 in libusb/os/libusb_usbfs.h

floe:
Right now, there is no stable/release version of libusb which works. The superspeed patch is already part of the official tree, but later changes in 1.0.19 messed things up again, so we simply checkout libusb-1.0 commit 51b10191033ca3a3819dcf46e1da2465b99497c2 which is known to work and build that.

Not ideal, and more or less a showstopper for "official" distribution packages, but unless someone manages to make libusb master work _and_ integrates the required changes in libusb _and_ creates a new release, there's not much we can do about it IMHO.

tikiking1:
So, I'm asking as the maintainer of the Arch Linux [`libfreenect2-git` AUR package](https://aur.archlinux.org/packages/libfreenect2-git). What do you think the best way to distribute libfreenect2 is? I think maybe putting the patched libusb in `/usr/lib/libfreenect2/libusb-1.0.so.0` would do it? Can you foresee any major problems from doing that?

tikiking1:
Okay, I just tried it, and `ldd` reports both `/usr/bin/Protonect` and [pyfreenect2](https://github.com/tikiking1/pyfreenect2)'s `_pyfreenect.so` using the patched version.

floe:
Now that I've been thinking about this for a little more, we should consider simply _statically_ linking to our patched libusb-1.0 - then the whole rpath mess would be avoided? /cc @xlz @larshg

xlz:
Yes.

floe:
[...]

PR #283, opened 2015-06-10
Link local libusb statically on Linux/Mac

xlz:
Following discussions in #281.

This builds libusb-1.0.a into libfreenect2.so. 

An issue with that is libfreenect2.so will export symbols that belong to libusb-1.0, then the users can't link libfreenect2 with anything that links with the system libusb-1.0 (symbol conflicts).

This works around the above issue by using GCC linker flag `--exclude-libs,libusb-1.0.a`. On Mac OSX, `ld64` does not support such operation, with only one solution: generate all symbols, remove `_libusb_*` symbols, and specify non-libusb symbols with `-exported_symbols_list,nolibusb.sym`.

If anyone has better approach, please comment below. Static linking on Windows is a seperate PR.


xlz:
@tikiking1 Does this PR work for your libfreenect2-git and pyfreenect2? Do they work correctly with static libusb?

tikiking1:
The Arch package builds fine, I'm updating it now. I don't actually have a Kinect 2 with me right now, but Protonect doesn't immediately segfault, so that's encouraging.

tikiking1:
And pyfreenect2 builds fine, but gets `ImportError: /usr/lib/libfreenect2.so: undefined symbol: udev_device_get_devnode` when I actually import it. Adding `udev` to the libraries section of `setup.py` doesn't fix it. It's a mystery?

xlz:
It's a bug. Thanks for the catch.

xlz:
The bug was caused by reversing the link order of libusb. Can you update and try again?

tikiking1:
OK, it works now (in terms of getting an error because I don't have a Kinect with me, rather than because of a linking issue)

tikiking1:
Is there a timeline for when this and #284 will get merged?

floe:
Not really, although I'm hoping I'll get around to it sometime next week. Problem is that there may still be some interference with #292.

xlz:
#292 is fairly easy to code up. The issue is a decision has to be made.

xlz:
I start to think embedded static libusb in libfreenect2 is a bad idea. The hack of messing with export symbols was not a good idea from the outset and can easily break. Besides, it is really not up to libfreenect2 to decide where to install libusb, or to embed it.

For packagers, I think the best option is to carry a separate libusb.so in a private directory of the package you want to release and change RPATH of libfreenect2 to that directory. `chrpath` can do that job post hoc. If you know the install directory before building, `-DCMAKE_INSTALL_RPATH=...` will do that too (not right now; need a quick patch).

xlz:
I have created a patch for specifying custom RPATH https://github.com/xlz/libfreenect2/commit/7a4708fa03d528ff9339c2c7034515e44861561a. It is expected to be part of a future PR soon.

PR #290, opened 2015-06-13
Fix compile error in libfreenect2.h

hanyazou:
I'd like to fix this syntax error in include/libfreenect2.h and include the file from my source code while it seems not to be used by anyone else.


floe:
It certainly won't hurt anyone to fix the compile error, but why do you want to include it anyway? This is a draft for a C API which noone ever implemented, and based on how libfreenect2 is constructed internally, it will probably be difficult to make this compatible...

PR #293, opened 2015-06-15
Build OpenCL program as soon as the OpenCL device is initialized

HenningJ:
Before this, the OpenCL program was built when the first frame arrives and the following frames were dropped, because building the program takes a while.

With this change, the OpenCL program is built before the device is started. Then, when the first frame arrives, it only needs to be initialized, which is quite fast.

When the `DepthPacketProcessor::Config` changes, the OpenCL program might need to be re-initialized or even be rebuilt (depending on which config values were changed), but only the config actually changed. This is handled here as well.

### DIFF
--- a/examples/protonect/src/opencl_depth_packet_processor.cpp
+++ b/examples/protonect/src/opencl_depth_packet_processor.cpp
@@ -138,10 +138,14 @@ public:
   cl::Buffer buf_filtered;
 
   bool deviceInitialized;
+  bool programBuilt;
   bool programInitialized;
   std::string sourceCode;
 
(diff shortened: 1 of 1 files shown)


floe:
Patch looks very sensible, on what platforms has this been tested so far?

HenningJ:
I just tested it on Windows

floe:
Could you merge master into your opencl-build branch again? I don't think this is really a merge conflict, there just have been some minor changes within the OpenCL depth packet processor.

HenningJ:
Done. It was just one little `throw`.

floe:
Finally got around to testing on Ubuntu 14.04, works beautifully. Merged.

PR #294, opened 2015-06-16
Various small fixes

laborer2008:


xlz:
Looks fine to me.

xlz:
The last commit is not small. And I do not agree with it.

It means each time a new optional feature is added, one or more new lines of `LIST(REMOVE_ITEM SRC ...)` needs to be created. This is worse than adding source files in the feature's own if block.

It is also prone to pick up unrelated files. In fact, `generate_resources.cpp` will be picked up after this commit, and Protonect will fail to build from source because `main()` is defined in `generate_resources.cpp`.

laborer2008:
xlz,
First of all, I haven't proposed that last commit yet. And it's not from this patchset.
I understand its drawbacks. But I just want to find (and propose) an improvement.
Existing code is not ideal either.

Secondly, The whole project tree seems to me deprecated now. It seems to me it was started
kind of spontaneously without attention to unimportant things such how to name folder or how deep
to place files. Seriously, why do we need to enter "examples" folder to get access to library code?
All mixed up.

Usually, a good project organized like this:

project root
|
|-(include)
|-sources
|-examples
|-tests
|-docs
|-utilities(scripts)
|-...

A good build system provides us with options. So we can choose whether we want to build just library or also examples(Protonect.cpp) and tests(test_opengl_depth_packet_processor.cpp) and documentation and utilities(generate_resources.cpp) and so on.

wiedemeyer:
Regarding the `SOURCE` things. Isn't it just used to show all project files in IDEs that support CMake? Why not simply add all (`*`) files that exist in `src` and `inlcude` as `SOURCES` to the project. Then there is no need to name each file. Also it will always show all files that exist in the IDE, not only the ones activated.

laborer2008:
wiedemeyer,

It's not about IDEs.
According to this
```
ADD_LIBRARY(freenect2 OBJECT ${SOURCES})
```
SOURCES lists all the files constituting library at the linking stage.

for *.cpp files we will get linking errors(several main function).

And besides these selectable options
```
OPTION(ENABLE_OPENCL "Enable OpenCL support" ON)
OPTION(ENABLE_OPENGL "Enable OpenGL support" ON)
```
(And I see CUDA and VA-API support is about to be added)

prevents us from using all the *.cpp files in any configuration.

laborer2008:
As for my last commit my bad I didn't create a standalone branch.
Didn't want to bring it here.

HenningJ:
But now that it's here...I agree with @xlz that it doesn't really make it better. I actually prefer to have the files explicitly listed (for example things git-bisect work better when the cmake files change when files are added/removed from the project).

But I also agree with @laborer2008 that the project structure is far from ideal.

laborer2008:
Removed that commit. For history it's [here](https://github.com/laborer2008/libfreenect2/commit/68fe7d180f5919250fdbe93bee58dd1c0efc3f9b)

wiedemeyer:
Ok, I mixed it up a bit. I thought is a seperate list of files. Because normally no headers need to be added there, the headers were only added to the list so that they appear in IDEs.

floe:
Best to keep PRs small and disjoint, so I will merge this (entirely sensible) one now and we'll (possibly) address the Cmake thing elsewhere.

PR #300, opened 2015-06-17
Extended name of libtubrojpeg for debian packaging

goldhoorn:


floe:
Wouldn't libturbojpeg.so.1 instead of .so.0 indicate a newer version of the package with different ABI?

goldhoorn:
On 17.06.2015 15:06, Florian Echtler wrote:
>
> Wouldn't libturbojpeg.so.1 instead of .so.0
>
that what the PR is. (regard naiming)
>
> indicate a newer version of the package with different ABI?
>
> —
>
I think debian marks their stable/unstable packages this way.
however i compiled successfully against this versions so no EABI 
incompatibility can occur.

Tested on debian stable/unstable/testing and ubuntu 14.04 and 15.04

floe:
OK, I'll merge it, but please add a comment that .so.1 is from Debian Jessie(?) and .so.0 is from Ubuntu 14/15. Otherwise, when the Ubuntu packages will update at some point, then .so.1 will point to a different versions on Ubuntu, possibly breaking things.

floe:
OK, thanks (you could also have pushed an additional commit to this PR, BTW).

goldhoorn:
Whos i thoug you merged _this_ one. sorry ;) too much prs from me going around ;)

xlz:
Debian's libturbojpeg{0,1} provide libturbojpeg.so. There is no need to work around it.

https://packages.debian.org/jessie/amd64/libturbojpeg1-dev/filelist
https://packages.debian.org/stretch/amd64/libturbojpeg0-dev/filelist
https://packages.debian.org/sid/amd64/libturbojpeg0-dev/filelist

goldhoorn:
I saw this on the homepae too, however our standard build server (no 
mods in the OS) tells me something different where without this patch 
the build is failing...

Best,
Matthias

xlz:
What exactly is the something different? Can we debug it?

Issue #295, opened 2015-06-16
Is file src/test_opengl_depth_packet_processor.cpp still necessary?


laborer2008:
According to CMakeLists.txt
```
IF(ENABLE_OPENGL)
  FIND_PACKAGE(GLFW3)
  FIND_PACKAGE(OpenGL)
  IF(GLFW3_FOUND)
    INCLUDE_DIRECTORIES(${GLFW3_INCLUDE_DIRS})

    LINK_DIRECTORIES(${GLFW3_LIBRARY_DIRS})
    LIST(APPEND LIBRARIES
      ${GLFW3_LIBRARIES}
      ${OPENGL_gl_LIBRARY}
    )
  SET(LIBFREENECT2_WITH_OPENGL_SUPPORT 1)
  LIST(APPEND SOURCES
    src/flextGL.c
    src/opengl_depth_packet_processor.cpp
  )
  
  LIST(APPEND RESOURCES
    src/shader/debug.fs
    src/shader/default.vs
    src/shader/filter1.fs
    src/shader/filter2.fs
    src/shader/stage1.fs
    src/shader/stage2.fs
  )
  ENDIF()
ENDIF(ENABLE_OPENGL)
```

File test_opengl_depth_packet_processor.cpp is not added to SOURCES and not linked. Seems it's deprecated some time ago and should be removed from the master branch.

laborer2008:
Now I see that this is another program. If this file is still important seems that another project should be created with its own CMakeLists.txt

gaborpapp:
See #169. It's not necessary for most users, but if the opengl packet processor is not working for some reason on your machine it is very useful for debugging.

xlz:
@gaborpapp Same question: is this still necessary? It seems the better way to debug OpenGL is to add lots of error checking.

xlz:
I will remove this file.

Issue #287, opened 2015-06-11
Libfreenect2 Documentation


jeppewalther:
Hi
Do there exist some form of documentation on Libfreenect2 internals other that the source code itself.

SebastianVarma:
Also, are there any differences between the libfreenect and libfreenect2 api's?

floe:
@jeppewalther I guess the closest thing we have to documentation is `protonect.cpp`, unfortunately.
@SebastianVarma Biggest difference: libfreenect is mostly a synchronous C API with a C++ wrapper, libfreenect2 is pure C++ and asynchronous with internal threading.

SebastianVarma:
@floe Are the function and variable names the same?

floe:
No - it's a quite different type of API, don't think it's comparable.

SebastianVarma:
@floe Is there any sort of guide or tutorial besides the documentation (protonect.cpp)?  Could one learn the library using libfreenect2/include/libfreenect2.h?

jeppewalther:
@floe Thanks for the info. BTW The documentation i was querying for was not a user guide. But documentation on how you have implemented the driver. My guess is that there should be a paper worth publishing with this information.

floe:
@jeppewalther that's not a bad idea - most of the actual reverse engineering was done by @joshblake @christiankerl and @sh0, so this is definitely a suggestion for those guys. 32C3 in Hamburg next winter might be a nice venue :-)

HenningJ:
But some more documentation on the API would be nice as well. Like: does the C-API in `libfreenect2/include/libfreenect2.h` actually relevant in any way? Or that the `FrameListener::onNewFrame()` callback has to delete the frame pointer once it's done with the frame. Or a better separation between the API and the internal implementation details. The actual API is just the stuff in `libfreenect2.hpp` and `frame_listener.hpp`, right?
You can figure all of this out by looking at the code, but it would be easier with some basic documentation. I'd be happy to help, but I'm not quite sure about any of it yet.

floe:
Yes, and that's a good point: `libfreenect2.hpp` and `frame_listener.hpp` should be installed by cmake to a separate include directory. If they are properly commented (which they absolutely are not at the moment), then that might be enough documentation for the first release...

xlz:
What's the standard way to generate doxygen docs?

[...]

PR #314, opened 2015-07-07
rename examples/protonect to src

floe:
This should probably be the very last PR to be merged for 0.1; alternative suggestions for the folder name welcome.


HenningJ:
Shouldn't protonect itself stay in `examples`? It's not really part of libfreenect2, but rather an example for it. :-)

christiankerl:
why not just put it in the root folder (as libfreenect)? src/src seems strange?!

floe:
@HenningJ @christiankerl good points, amended.

HenningJ:
Two more things (taken from [issue #287](https://github.com/OpenKinect/libfreenect2/issues/287#issuecomment-111414161)):
* remove `include/libfreenect2.h` ? Since it's not actually relevant/used anywhere, it only adds confusion.
* separate the public API from the implementation by moving the headers to a different folder. That's a bit more complicated (but it would be nice for the 0.1 release), because it's not clearly separated yet. `libfreenect2.hpp` and `frame_listener.hpp` are the API, but you'd also need the `PacketPipelines`. And oy crouse the registration should be part of the API as well.

Also in [PR #294](https://github.com/OpenKinect/libfreenect2/pull/294#issuecomment-112764701) @laborer2008 proposes a slightly different structure. That would provide even better separation, like moving `generate_resources.cpp` to `tools/` and `test_opengl_depth_packet_processor.cpp`to `tests/`, making it even clearer, which files are actually part of libfreenect2 and which files are built around it to support it. Just some food for thought.

xlz:
libfreenect2/libfreenect2 still seems strange. 

I would like to move `examples/protonect` into the root directory like this:

```
cmake_modules/
include/libfreenect2/
src/
data/
    11to16.bin
    xTable.bin
    zTable.bin
tools/
    generate_resources.cpp
tests/
    test_opengl_depth_packet_processor.cpp
    Protonect.cpp (this is also used as a basic test case)
CMakeLists.txt
freenect2.cmake.in
freenect2.pc.in
```

By the way, in README.md please also change the build instructions to `mkdir build && cd build && cmake ..` which is the norm using CMake that does not pollute the source directory. (I thought everyone knows this, but it seems otherwise from many bug reports.)

wiedemeyer:
+1 @xlz (especially the build folder part)
But Protonect could also go into tools, if you look at it as a viewer, or in examples as an example on how to use libfreenect2.

xlz:
@wiedemeyer I don't mind either way.

@laborer2008 Now that tests and tools are moved out of src, you may be interested in reviving the attempt in #295 of using glob to list source files instead of explicitly listing them in CMakeLists.txt. 

A potential approach of modularization would move non-core modules into their subdirectories in `src/`:

```
src/opengl/
    shader/
    flextGL.c
    flextGL.h
    opengl_depth_packet_processor.cpp
src/opencl/
    opencl_depth_packet_processor.cl
    opencl_depth_packet_processor.cpp

In the future:
src/cuda/
src/vaapi/
src/tegra/
```

Each module could carry their own CMakeLists.txt, and then the root CMakeLists.txt could automatically adds all subdirectories. I realize more refactoring is required for complete modularization. I may open a new PR for this in the future.

laborer2008:
xlz, Yes, of course. I see you understand my idea completely. But I myself would prefer to wait till this changeset is merged and those CMakeLists.txt someone create. Don't want to mess things here up.

Also I want to notice:
- include/freenect2 looks for me better than include/```lib```freenect2, especially on Windows. Even if library is called "libfreenect2". Just look on the other opensource libraries.
- Paths like src/```opencl```/depth_packet_processor.cl look  for me better than src/```opencl```/```opencl_```depth_packet_processor.cl
- Shouldn't Protonect.cpp be called protonect.cpp as other files? If someone wants 'Protonect.exe' for executable on Windows (because it's conventional) it can be done in CMakeLists.txt

xlz:
@laborer2008 These cosmetic name changes are likely to break things. I want to avoid them.

floe:
@xlz I think your initial suggestion is the most sensible one which requires the least amount of renames/edits. I'll close this one and create a new PR, tracking all these renames across the other PRs would be somewhat difficult.

PR #315, opened 2015-07-09
fixed memory leak in openDevice

wiedemeyer:
There is a memory leak in `Freenect2::openDevice`.
If it fails before [line 740](https://github.com/OpenKinect/libfreenect2/blob/master/examples/protonect/src/libfreenect2.cpp#L740) the PacketPipeline will not be deleted. It could also happen when passing a wrong serial number.
If it fails after [line 740](https://github.com/OpenKinect/libfreenect2/blob/master/examples/protonect/src/libfreenect2.cpp#L740) the PacketPipeline will be deleted by the destructor of `Freenect2DeviceImpl`.

I added the necessary delete instructions. We should note somewhere in the documentation that libfreenect2 takes over control of the pipeline object (including destruction) and that it should never be used or deleted by user code afterwards.

### DIFF
--- a/examples/protonect/include/libfreenect2/libfreenect2.hpp
+++ b/examples/protonect/include/libfreenect2/libfreenect2.hpp
@@ -106,6 +106,11 @@ public:
   std::string getDeviceSerialNumber(int idx);
   std::string getDefaultDeviceSerialNumber();
 
+  /* Important note:
+   * After passing a PacketPipeline object to libfreenect2 do not use or free the object,
+   * libfreenect2 will take care. If openDevice fails the PacketPipeline object will get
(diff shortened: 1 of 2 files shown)


floe:
For the moment, could you just add a comment in `libfreenect2.hpp`? Then I can merge it in one go.

wiedemeyer:
done.

HenningJ:
While we're on the subject of memory mangement, you could also add a comment on the `FrameListener::onNewFrame()` method that the `frame` pointer must be release in there.

wiedemeyer:
There are probably some more parts where comments are useful. But this is out of the focus of this PR.

floe:
OK, merged. We'll probably need another comments/docs PR anyway.

PR #317, opened 2015-07-10
allow supplying an external Frame for the depth buffer

floe:
The depth buffer created by `Registration::apply` basically is already an upsampled, color-aligned depth image. When an external `Frame` object is used as this buffer, it can be re-used for filtering the high-res color image.


xlz:
Looks quite simple and unintrusive. Just merge it.

Issue #165, opened 2015-02-12
'OpenCL/cl.hpp' file not found


jessekirbs:
Running a Mac with OSX 10.9.5. Everything goes smooth until 'make && make install'. I get this error:

```
/Users/jessekirberger/Documents/Kinect/libfreenect2/examples/protonect/src/opencl_depth_packet_processor.cpp:43:10: fatal error: 
      'OpenCL/cl.hpp' file not found
#include <OpenCL/cl.hpp>
         ^
1 warning and 1 error generated.
make[2]: *** [CMakeFiles/freenect2.dir/src/opencl_depth_packet_processor.cpp.o] Error 1
make[1]: *** [CMakeFiles/freenect2.dir/all] Error 2
make: *** [all] Error 2
```

Any ideas? I don't have a folder called 'OpenCL' anywhere, but I do have the OpenCL.frameworks folder.

larshg:
What GPU are you using and driver/Opencl sdk?

jessekirbs:
@larshg NVIDIA GeForce GT 650M 512 MB. Not really sure what OpenCL SDK I'm using.. how do I check that? How would I disable the OpenCL and use OpenGL? I'm fairly new to all this. Thanks for the reply.

larshg:
If you use CMake GUI you can go under ENABLE and deselect ENABLE_OPENCL.
If you use CMake in a command prompt I'm not entirely sure how it should be passed - maybe like this: cmake -DENABLE_OPENCL=OFF - or something like it :)

Otherwise you could try add the following:

#define CL_USE_DEPRECATED_OPENCL_1_1_APIS
+#include <CL/cl.h>
+#undef CL_VERSION_1_2

in opencl_depth_packet_processor.cpp - instead of the #include <OpenCL/cl.hpp>

As far as I know - Nvidia doesn't support the CL version 1.2 yet and on my machine with nVidia, the cl.hpp file wasn't there.

I haven't tested the part about OpenCL includes, so I might be totally off track here :)

jessekirbs:
@larshg You are legend, my friend. 'cmake -DENABLE_OPENCL=OFF' did the trick and it installed correctly. Thank you so much. Now it's having trouble finding the Kinect. Does it ever end? 

```
[Freenect2Impl] enumerating devices...
[Freenect2Impl] 13 usb devices connected
[Freenect2Impl] found 0 devices
[Freenect2Impl] requested device 0 is not connected!
[...]

PR #318, opened 2015-07-11
Use newer OpenCL include path to save cl.hpp

hanyazou:
Latest Mac OS X 10.10 does not have /System/Library/Frameworks/OpenCL.framework/Versions/A/Headers/ folder. The folder seems to be deprecated. This change handle newer location in depends/install_mac.sh.

(related to #165 and a39b0e0)

### DIFF
--- a/depends/install_mac.sh
+++ b/depends/install_mac.sh
@@ -2,12 +2,14 @@
 
 cd `dirname $0`
 DEPENDS_DIR=`pwd`
+OS_VER=`sw_vers -productVersion | sed 's/^\([0-9]*\)\.\([0-9]*\).*$/\1.\2/'`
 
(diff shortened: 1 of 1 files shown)


depends/install_mac.sh
```diff
@@ -2,12 +2,14 @@
 
 cd `dirname $0`
 DEPENDS_DIR=`pwd`
+OS_VER=`sw_vers -productVersion | sed 's/\([0-9]*\).\([0-9]*\).*$/\1.\2/'`
```
floe: Should there be a backslash before the first dot in the RegExp?

hanyazou:
@floe You are right. Fixed. Thank you.

PR #326, opened 2015-07-15
make sure data pointer in Frame object is 64-byte aligned 

floe:
When using stuff like SSE/AVX, the data needs to be aligned to 32- (or 64-) byte boundaries. This makes sure that the data pointer in a Frame object conforms to that alignment without needing stuff like `_aligned_malloc`. I don't think this can break anything, but opinions are welcome @christiankerl @xlz

### DIFF
--- a/examples/protonect/include/libfreenect2/frame_listener.hpp
+++ b/examples/protonect/include/libfreenect2/frame_listener.hpp
@@ -34,8 +34,9 @@
 namespace libfreenect2
 {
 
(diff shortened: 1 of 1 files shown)


christiankerl:
make rawdata private or protected?

floe:
@christiankerl likely a good idea, would require changing from `struct Frame` to `class Frame`. Will also include that.

xlz:
Fine by me now.

Issue #330, opened 2015-07-20
Bug: VS 2012 (Express) build failures


AliShug:
I've had a few issues with getting the build to go through using Visual Studio Express 2012 - I would have changed environments, but didn't have that option:

* Trying to link with OpenCV 3.0 resulted in a notfound error in the first stage of the cmake build. I don't know if this is expected or if I screwed up the installation, either way switching to my existing 2.4 directory seemed to solve this.

* Libusb wasn't correctly linked. The script tried to link:
  `C:/Users/.../depends/libusb/x64/Release/lib` (the directory)
  When it needed:
  `C:/Users/.../depends/libusb/x64/Release/lib/libusb-1.0.lib`

* I had to manually specify the TurboJPEG include and library variables. Not sure what went wrong here, the variables appeared but were set with error values.

* My version of MSVC has terrible C99 support, so doesn't define `roundf`. I had to insert a replacement function into registration.cpp (Registration::Registration uses it a few times)

xlz:
> OpenCV 3.0

OpenCV 3.0 has API incompatibilities. We are in the process of removing OpenCV altogether. Please use OpenCV 2.x for the moment.

> Libusb wasn't correctly linked

Need CMake output for this.

> had to manually specify the TurboJPEG
> variables appeared but were set with error values.

What are the wrong values? Again, CMake output please. 

> My version of MSVC has terrible C99 support, so doesn't define roundf

@wiedemeyer I think I have seen one of your patch doing rounding with `(int)(x + 0.5f)`. Shouldn't it already replace all `roundf()`?

AliShug:
> Please use OpenCV 2.x for the moment.

No problem, but it might be worth mentioning that in the Readme?

I'm away from my workstation at the moment but I'll try a clean re-build when I get the chance, and post the CMake output here. Anyway, it seemed like it couldn't find the TurboJPEG include and library directories from the provided root, but that might well have been a screw-up on my part; the LibUSB error was much stranger.

There's `roundf()` use in [registration.cpp](https://github.com/OpenKinect/libfreenect2/blob/master/examples/protonect/src/registration.cpp#L244-L258) at lines 244, 245 and 258

AliShug:
Okay I've tried again from scratch, and it looks like the build problems were mostly errors on my part; I accidentally built the static libusb library. Don't know what happened with TurboJPEG though.

Apologies for the red herrings - that said, the Readme's build instructions could do with fixing:

```bash
cd example\protonect
[...]

PR #339, opened 2015-07-22
Quick windows docs update, address #330

xlz:
Fix a typo.
Only OpenCV 2.x is supported for Windows.


larshg:
I'm running with opencv 3.0 (recent master) and it works fine...

xlz:
This PR is almost too trivial. OpenCV is getting removed so why care?

I suppose users are smart enough to spot the missing s in `example\protonect`.

PR #341, opened 2015-07-22
Exit on opengl errors

larshg:
Added error message if creation of flextgl, glfw or glfwwindow fails.


floe:
Thanks, a very sensible idea; merged.

PR #345, opened 2015-07-23
Remove `roundf()` use from Registration

AliShug:
Replaces use of `roundf()` function in registration.cpp with `(int)(x +
0.5f)` to allow compiling on older versions of MSVC.


floe:
@wiedemeyer any comments - this should work transparently AFAICT?

PR #351, opened 2015-07-27
Correct find_scrpipt for libusb

goldhoorn:
The DEPENDS is only set for a local installation.
Otherwise the system (global) one should used.
Furthermore the check if libusb was actually found
(even reuqired) was broken

### DIFF
--- a/examples/protonect/cmake_modules/FindLibUSB.cmake
+++ b/examples/protonect/cmake_modules/FindLibUSB.cmake
@@ -8,8 +8,14 @@
 #  LibUSB_LIBRARIES:  the libraries
 
 IF(PKG_CONFIG_FOUND)
-  SET(ENV{PKG_CONFIG_PATH} "${DEPENDS_DIR}/libusb/lib/pkgconfig")
-  PKG_CHECK_MODULES(LibUSB libusb-1.0)
+  IF(${DEPENDS_DIR}) #Otherwise use System pkg-config path
+    SET(ENV{PKG_CONFIG_PATH} "${DEPENDS_DIR}/libusb/lib/pkgconfig")
+  ENDIF()
+  IF(LibUSB_FIND_REQUIRED)
+    PKG_CHECK_MODULES(LibUSB REQUIRED libusb-1.0)
+  ELSE()
+    PKG_CHECK_MODULES(LibUSB libusb-1.0)
+  ENDIF()
   RETURN()
 ENDIF()
 


xlz:
> The DEPENDS is only set for a local installation. Otherwise the system (global) one should used.

This assertion is wrong. From pkg-config docs:

```
PKG_CONFIG_PATH
       A colon-separated (on Windows, semicolon-separated)  list  of  directories  to
       search  for  .pc  files.   The default directory will always be searched after
       searching the path; the default  is  libdir/pkgconfig:datadir/pkgconfig  where
       libdir  is the libdir for pkg-config and datadir is the datadir for pkg-config
       when it was installed.
```

> Furthermore the check if libusb was actually found (even reuqired) was broken

Where is it broken? It seems the second part only changes this
```
if()
  xxx
  return()
endif()
yyy
```
to
```
if()
   xxx
else()
   yyy
endif()
```
which does nothing.

goldhoorn:
@xlz: 

regarding the assertion, it overrides the pkg_config path if the user adds the pkg-config path in his envorment this got overridet here. It has had a impact for me

regarding the if/then/else, i forgot to remove the return *fixed* the last two lines are important

xlz:
> The default directory will always be searched after searching the path; the default  is  libdir/pkgconfig:datadir/pkgconfig

Whatever the user specifies, the default will always be searched. If `DEPENDS_DIR` is not set, the default is searched. I don't see how it is possible for pkg-config to not provide a valid result here if you have installed pkg-config correctly even given invalid `PKG_CONFIG_PATH`.

>  i forgot to remove the return fixed the last two lines are important

No, the `RETURN` here is a catchall phrase for all Linux/Mac processing. The rest is for Windows.

goldhoorn:
@xlz: 
SET(ENV{PKG_CONFIG_PATH} "${DEPENDS_DIR}/libusb/lib/pkgconfig")
This line overrides the setting if it exists, therefore pkg_config behaved differently

Scenario:
we install prefix in ~/dev/install
if you have the env PKG_CONFIG_PATH=~/dev/install/lib/pkgconfig
this got overridden in the old setup to a empty string if depens dir is not set.

the return returns for linux/mac processing but the script then does NOT check if the library was actually found by pkg-config, therfore the last two lines needs to be executed.

xlz:
```diff
-  SET(ENV{PKG_CONFIG_PATH} "${DEPENDS_DIR}/libusb/lib/pkgconfig")
+  SET(ENV{PKG_CONFIG_PATH} "$ENV{PKG_CONFIG_PATH}:${DEPENDS_DIR}/libusb/lib/pkgconfig")
```

This solves your problem.

The last two lines are not designed to be used for Linux/Mac. `pkg_check_modules` should report error, not the extra checking.

goldhoorn:
This then adds a empty path (/libusb/lib/pkgconfig) to the PKG_CONFIG_PATH which i don't recomment

The pkg_check_modules only thecks if REQUIRED is passed as last argument, which is not default for this scipt

xlz:
Now I guess your use case is when you remove `DEPENDS_DIR` from the main CMakeLists.txt? 

I'm fine with this part is that is what you want:
```diff
-  SET(ENV{PKG_CONFIG_PATH} "${DEPENDS_DIR}/libusb/lib/pkgconfig")
+  if(${DEPENDS_DIR}) #Otherwise use System pkg-config path
+    SET(ENV{PKG_CONFIG_PATH} "${DEPENDS_DIR}/libusb/lib/pkgconfig")
+  endif()
```

The `RETURN` here is an important separation of the build system between Linux and Windows. Please do not remove `RETURN` and use `REQUIRED` for pkg_check_modules, which reports error if not found.

goldhoorn:
ok the first part is as requested.

but the recond part then enfoces libusb to be found, while the caller of this script does not enforce if find_package can be called without REQUIRED in general.

xlz:
libusb is always required. There is no need for this generalization.

goldhoorn:
> libusb is always required. There is no need for this generalization.

I don't agree here if someone will copy this script (which might happen) it should as correct as possible.
However i get it now correctly working and updated the PR again.

xlz:
You need properly parse `REQUIRED` out of `find_package` arguments.
```cmake
  foreach(_pkg ${ARGN})
    if (_pkg STREQUAL "REQUIRED")
      set(LibUSB_FIND_REQUIRED 1)
    endif ()
  endforeach()
```

goldhoorn:
I assume the required flag is already set, i checked it in my current commit and it works
see 

http://www.cmake.org/cmake/help/v3.0/command/find_package.html

xlz:
This is correct per cmake docs. Fine to merge by me.

floe:
Thanks for the thorough discussion, merged.

PR #357, opened 2015-08-07
Corrected handling of DEPENDS_DIR and extended description of it

goldhoorn:


xlz:
```diff
-  SET(ENV{PKG_CONFIG_PATH} "${DEPENDS_DIR}/libusb/lib/pkgconfig")
+  SET(ENV{PKG_CONFIG_PATH} "$ENV{PKG_CONFIG_PATH}:${DEPENDS_DIR}/libusb/lib/pkgconfig")
```
was perfectly fine.

I didn't know why you want to do this. So you have libusb installed somewhere else than in $DEPENDS_DIR, OK. You can just `export PKG_CONFIG_PATH=xxx` as you did and pkg-config will properly use your priority, with `DEPENDS_DIR` in CMakeLists.txt untouched, and the expression still being valid.

> This then adds a empty path (/libusb/lib/pkgconfig) to the PKG_CONFIG_PATH which i don't recomment

Why does `DEPENDS_DIR` have to be empty in the first place? `$ENV{PKG_CONFIG_PATH}` takes priority in the above diff.

goldhoorn:
@xlz: what do you talking about?, your snippets have nothing to do with this PR?
And i already explained to you why a simple PATH extension is NOT what it is correct here...

xlz:
@goldhoorn 

You explained adding "a empty path (/libusb/lib/pkgconfig) to the PKG_CONFIG_PATH" to PKG_CONFIG_PATH is not correct. But you did NOT explain why ${DEPENDS_DIR} must be an empty path in the first place.

DEPENDS_DIR has always been an internal hardcoded variable in CMakeLists.txt. You were doing things assuming it is set to empty, and then this PR will force users to specify DEPENDS_DIR to "false" to avoid breaking pkg-config with your assumption, but this is totally unnecessary as I explained above.

goldhoorn:
szenario 1. User uses external system-dependancies
DEPENDS_DIR is set so libfreenect2_SOURCE_DIR/../../depends (so valid)
this results that the env got extended to:

SET(ENV{PKG_CONFIG_PATH} " libfreenect2_SOURCE_DIR/../../depends/libusb/lib/pkgconfig")

which is then maybe a empty dir so cmake find's libusb in the system. 

Szenario 2:
depends installed, path get added as above, everything is fine....

Szenario 3: (my example, user has libusb ANYWHERE installed and has set PKG_CONFIG_PATH on his own.

Then IF depends dir not set fo false, the build system will override the ENV to the same value than in 1 or 2. 


Your suggest now a usage of:
``` SET(ENV{PKG_CONFIG_PATH} "$ENV{PKG_CONFIG_PATH}:${DEPENDS_DIR}/libusb/lib/pkgconfig")``` which would result that a maybe invalid path got added. Szenario one is already wrong because the env should not get added at all. I don't want to break behaviour even it is bad, so i done at least szenario 3 the correct way.

xlz:
> Your suggest now a usage of: SET(ENV{PKG_CONFIG_PATH} $ENV{PKG_CONFIG_PATH}:${DEPENDS_DIR}/libusb/lib/pkgconfig") which would result that a maybe invalid path got added.

This results in `PKG_CONFIG_PATH` being `your_custom_path:libfreenect2_SOURCE_DIR/../../depends/libusb/lib/pkgconfig`, and pkg-config will search first in `your_custom_path` and find your custom libusb. How is this invalid?

goldhoorn:
Because adding a "nonsense" path path is invalid.
It makes absolutly no sense to "automatically" adding suggested paths in 
case the user specifies them.
overrideing the ENV here is a bad thing in general but okay, this 
project does it this way. but for extended use-cases i want to prevent 
this behaviour.

xlz:
If it always works, why is it "invalid"? Some imaginary concerns that does not ever happen? Well, it makes great sense to me to appropriately use environment variables to control pkg-config behaviors.

goldhoorn:
On 10.08.2015 17:40, Lingzhu Xiang wrote:
>
> If it always works, why is it "invalid"? Some imaginary concerns that 
> does not ever happen? Well, it makes great sense to me to 
> appropriately use environment variables to control pkg-config behaviors.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/OpenKinect/libfreenect2/pull/357#issuecomment-129499460>.
>


It is wrong because it will adds paths that might or not might be 
invalid. There might be a installation left.
Adding things because the "do not break" (in one scenario) does not mean 
that they are not invalid, or kind of "bad smell" 
(https://en.wikipedia.org/wiki/Code_smell)
I would add all windows paths, all common linux paths, and also 
$HOME/install to the package_config path, but this does not mean that it 
is the correct way of doing it.

xlz:
FindXXX.cmake is the exact place to garner code smell (collect all kinds of defaults and fallback paths) so that the rest of the place is clean. You can look at other FindXXX.cmake files and see how they collect default and fallback paths.

If a particular pkg-config path is used by the project by default, it is very correct to append it to the end of `PKG_CONFIG_PATH` as a fallback. In that case, most users will need `libfreenect2_SOURCE_DIR/../../depends/libusb/lib/pkgconfig` as the default, and your case is a special case.

xlz:
@floe Let's merge this first.

This PR does fix a bug.

floe:
Sorry for the delay, merged.

PR #360, opened 2015-08-11
Add postfix to have both debug and release libraries.

larshg:
I don't know if this is used on linux/mac?


xlz:
Linux packaging usually builds things with RelWithDebInfo and strips debug info out to a separate dbg package.

I think this is fine as you isolate this config to Windows only.

PR #361, opened 2015-08-13
Removes opencv dependency

fran6co:
Rebased and squashed #261


floe:
@larshg OK if I merge this one instead of your original PR?

fran6co:
I squashed some changes into larshg's commit for standardizing the timer and the opencv header.

fran6co:
Wouldn't it be better to remove opencv dependency for good?

larshg:
@floe sure, as I don't have much time right now it's nice someone continues the work :-)

fran6co:
Removed opencv dependency, the only problem would be ```test_opengl_depth_packet_processor.cpp```. But I think it needs it's own cmake rules

floe:
Could you add division by 4500.0f in the grayfragment shader, then I think we're good to go.

fran6co:
I don't have the kinect at hand at the moment to test it out. Could you check if it's okay? https://github.com/OpenKinect/libfreenect2/pull/361/files#diff-8ca25d621f26a3215be9537b77208b46R59

xlz:
The last commit introduces hard dependency on C++11. As discussed before in #261, we want to avoid this.

Can you add a wrapper like this?
```cpp
double getTime()
{
#if __cplusplus >= 201103L
  //not tested:
  static auto start = std::chrono::high_resolution_clock::now();
  auto end = std::chrono::high_resolution_clock::now();
  auto duration = end - start;
  return std::chrono::duration_cast<std::chrono::duration<double>>(duration).count();
#elif LIBFREENECT2_WITH_OPENGL_SUPPORT
  return glfwGetTime();
#else
  return 0;
#endif
}
```

fran6co:
Made a new Timer class, when #309 is merged it can be moved to the logging source files.

xlz:
The rest looks fine to me.

fran6co:
Now it's using the full resolution of chrono.

xlz:
Tests OK on Intel/Linux/OpenCL.

floe:
Alright, last big one. I'm pretty certain it will still break in some contexts, but if we don't merge it now, this release will never happen :-)

PR #364, opened 2015-08-20
Logging refactoring continued

xlz:
Continuing from #309.

I changed the logging API to `libfreenect2::{get,set}GlobalLogger()` (instead of `setLogger()`) in case in the future we want the non-global logger interface again. The rationale of reverting to a global logger is described in this commit "Convert to a global static logger" and in the discussion in #309.

libfreenect2 still manages the memory of the logger passed in because it needs access to the logger until the very end in all kinds of destructors.


xlz:
@floe

This PR is complete. I have tested (Linux/Ubuntu) all combinations of `cmake .. -DENABLE_OPENGL={ON,OFF} -DENABLE_CXX11={ON,OFF} && ./Protonect {-noviewer,} {gl,cl,cpu}`. All work as expected.

xlz:
Rebased a little bit.

xlz:
The last fix needs more investigation.

xlz:
So the issue is quite complicated.

At first it was whenever the OpenGL processor prints errors like this `std::cerr << "1" << "2"; exit(-1);`, only the first "1" is printed, with the rest lost. I checked the stack at that moment, it looked like this:

```
#0  0x00007ffff65f64d4 in __GI___pthread_mutex_lock (mutex=0x7ffff7ffd968 <_rtld_global+2312>) at ../nptl/pthread_mutex_lock.c:114
#1  0x00007ffff7ddb150 in tls_get_addr_tail (ti=0x7ffff7543718, dtv=0x7ffff7fd8090, the_map=0x7ffff7fe89f8) at dl-tls.c:722
#2  0x00007ffff72b498c in __cxxabiv1::__cxa_get_globals () at ../../../../src/libstdc++-v3/libsupc++/eh_globals.cc:62
#3  0x00007ffff72b48b9 in std::uncaught_exception () at ../../../../src/libstdc++-v3/libsupc++/eh_catch.cc:139
#4  0x00007ffff72ebae2 in ~sentry (this=0x7fffffffce00, __in_chrg=<optimized out>) at /build/buildd/gcc-4.8-4.8.4/build/x86_64-linux-gnu/libstdc++-v3/include/ostream:431
#5  std::__ostream_insert<char, std::char_traits<char> > (__out=..., __s=__s@entry=0x40849a "device serial: ", __n=15) at /build/buildd/gcc-4.8-4.8.4/build/x86_64-linux-gnu/libstdc++-v3/include/bits/ostream_insert.h:112
#6  0x00007ffff72ebec7 in std::operator<< <std::char_traits<char> > (__out=..., __s=0x40849a "device serial: ") at /build/buildd/gcc-4.8-4.8.4/build/x86_64-linux-gnu/libstdc++-v3/include/ostream:535
```

The `std::ostream::sentry`'s destructor flushes the stream and outputs the content. During that it also checks `std::uncaught_exception()`. Nothing is wrong with that. In `std::uncaught_exception()` `__cxa_get_globals()` accesses a [static thread local variable](https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/libsupc%2B%2B/eh_globals.cc#L51). Thread local access `__tls_get_addr()` calls into `tls_get_addr_tail()`.

The commit "Fix exception handling in LogMessage destructor" removed usage of `operator<<()` for outputting, and outputting will no longer access thread local variables. But this only works around the problem.

I tried adding some `std::cerr <<` in `Protonect.cpp`, and it behaved like the above again. And actually the main thread was hanging, in a deadlock or something. This means in the past, all `std::cerr <<` in the codebase will probably cause the calling thread to hang. That may explain some of the OpenGL black screen bugs.

The only other access to thread-local variables happens in
```
#0  __tls_get_addr (ti=0x7ffff65eafb0) at dl-tls.c:770
#1  0x00007ffff63e6b8a in _initClIcd () at ocl_icd_loader.c:560
#2  0x00007ffff7dea13a in call_init (l=<optimized out>, argc=argc@entry=2, argv=argv@entry=0x7fffffffe388, env=env@entry=0x7fffffffe3a0) at dl-init.c:78
#3  0x00007ffff7dea223 in call_init (env=<optimized out>, argv=<optimized out>, argc=<optimized out>, l=<optimized out>) at dl-init.c:36
#4  _dl_init (main_map=0x7ffff7ffe1c8, argc=2, argv=0x7fffffffe388, env=0x7fffffffe3a0) at dl-init.c:126
#5  0x00007ffff7ddb30a in _dl_start_user () from /lib64/ld-linux-x86-64.so.2
```

According to https://bugzilla.redhat.com/show_bug.cgi?id=1219646, ocl-icd implementation was fundamentally flawed (before 2.2.4) in its use of thread local storage in the library constructor. If I disable OpenCL, or update ocl-icd-libopencl1 to 2.2.7-2 (Ubuntu 15.10 Wily Werewolf), the problem goes away.

Other bug reports of the same issue:
https://bugzilla.redhat.com/show_bug.cgi?id=1223055
https://bugzilla.redhat.com/show_bug.cgi?id=661676
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54018

According to GCC docs, "No static initialization may refer to the address of a thread-local variable." It looks like the case in these reports.

Now that many users are stuck with Ubuntu 14.04 with the old version of ocl-icd, a work around has to be provided. Suppose no further access to any thread local variable is allowed under the buggy ocl-icd. That means no access to `std::uncaught_exception` (`std::ostream::sentry::~sentry()` and `std::condition_variable_any`), and no access to anything related to exception:

```
__cxxabiv1::__cxa_tm_cleanup
__cxxabiv1::__cxa_rethrow
std::current_exception
__cxxabiv1::__cxa_allocate_exception
__cxxabiv1::__cxa_allocate_dependent_exception
__cxa_current_exception_type
__cxxabiv1::__cxa_begin_catch
__cxxabiv1::__cxa_end_catch
```

To avoid `std::uncaught_exception()`, only needs to disable automatic flushing:
```
std::nounitbuf(std::cerr);
```

It is harder to avoid exceptions. `-fno-exceptions` will disable c++ exceptions. Several places in the OpenCL processor and tinythread/tinythread.cpp need fixing. Probably use a macro LIBFREENECT2_NOEXCEPTION.

There are some other issues with OpenCL (ocl-icd version detection and NOEXCEPTION setup, CMake cleanup, macro detection, embedding a cl.hpp because newer opencl-headers no longer has one). In terms of PRs I'll probably address all OpenCL issues in a new one.

xlz:
The above analysis means the last fix is unneeded. For platforms without the buggy ocl-icd, std::cerr will work as expected instead of replacing it with stderr. The next step is to remove the last fix and add a commit of platform specific check for buggy ocl-icd.

floe:
AFAICT, to keep the changesets down to reasonable size, the PRs should now be merged in order #364 -> #368 -> #376 - is that correct?

xlz:
I think #368 will be mergeable after merging #364.

#376 probably needs some review (and fixing conflict).

xlz:
Rebased to master.

floe:
Thanks for your work. I've tested this on my machine, and everything works fine, so let's merge it.

PR #368, opened 2015-08-22
Intel Mesa OpenGL bug fixes and cleanup

xlz:
This is rebased on #364. Commits of this PR begin from "changed minimal opengl version to 3.1".

This PR fixes all bugs on Intel GPU. This PR supersedes #352.

Performance comparison on i7-4600U @ 2.10GHz:

```
[Info] [OpenGLDepthPacketProcessor] avg. time: 21.3259ms -> ~46.8913Hz
[Info] [OpenCLDepthPacketProcessor] avg. time: 16.9965ms -> ~58.8356Hz
```

### DIFF
--- a/examples/protonect/src/opengl_depth_packet_processor.cpp
+++ b/examples/protonect/src/opengl_depth_packet_processor.cpp
@@ -32,7 +32,8 @@
 #include <GLFW/glfw3.h>
 
 #include <fstream>
-
+#include <string>
+#include <map>
 
 #include <stdint.h>
 
@@ -117,6 +118,8 @@ bool loadBufferFromFile(const std::string& filename, unsigned char *buffer, size
 
 struct ShaderProgram : public WithOpenGLBindings
 {
+  typedef std::map<std::string, int> FragDataMap;
+  FragDataMap frag_data_map_;
   GLuint program, vertex_shader, fragment_shader;
 
   char error_buffer[2048];
@@ -142,6 +145,11 @@ struct ShaderProgram : public WithOpenGLBindings
     int length_ = src.length();
     fragment_shader = gl()->glCreateShader(GL_FRAGMENT_SHADER);
     gl()->glShaderSource(fragment_shader, 1, &src_, &length_);
+  }
+
+  void bindFragDataLocation(const std::string &name, int output)
+  {
+    frag_data_map_[name] = output;
   }
 
   void build()
@@ -171,6 +179,11 @@ struct ShaderProgram : public WithOpenGLBindings
     program = gl()->glCreateProgram();
     gl()->glAttachShader(program, vertex_shader);
     gl()->glAttachShader(program, fragment_shader);
+
+    for(FragDataMap::iterator it = frag_data_map_.begin(); it != frag_data_map_.end(); ++it)
+    {
+      gl()->glBindFragDataLocation(program, it->second, it->first.c_str());
+    }
 
     gl()->glLinkProgram(program);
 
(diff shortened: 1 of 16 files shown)


examples/protonect/src/opengl_depth_packet_processor.cpp
```diff
@@ -347,10 +424,10 @@ struct OpenGLDepthPacketProcessorImpl : public WithOpenGLBindings
   Texture<U16C1> input_data;
 
   Texture<F32C4> stage1_debug;
-  Texture<F32C3> stage1_data[3];
+  Texture<F32C4> stage1_data[3];
```
RyanGordon: Just curious, what was the particular effect of the old value?

examples/protonect/src/opengl_depth_packet_processor.cpp
```diff
@@ -117,31 +124,81 @@ bool loadBufferFromFile(const std::string& filename, unsigned char *buffer, size
...
   ShaderProgram() :
     program(0),
     vertex_shader(0),
-    fragment_shader(0)
+    fragment_shader(0),
+    mesa_checked(false)
+  {
+  }
+
+  bool checkMesaVersion()
```
RyanGordon: Would this be better named as isBuggyMesaVersion or similar?

examples/protonect/src/opengl_depth_packet_processor.cpp
xlz: The effect is the texture can't be attached to FBO. With `INTEL_DEBUG=fbo,...` you get `FBO incomplete: Unsupported HW texture/renderbuffer format attached: MESA_FORMAT_RGB_FLOAT32`

xlz: Original diagnosis was in #352.

examples/protonect/src/opengl_depth_packet_processor.cpp
xlz: I have no preference for any name. I'll change it in the next update to the PR.

examples/protonect/src/shader/stage1.fs
```diff
@@ -106,6 +104,9 @@ void main(void)
   vec2 ab1 = processMeasurementTriple(uv, P0Table1, 3, Params.ab_multiplier_per_frq.y, saturated.y);
   vec2 ab2 = processMeasurementTriple(uv, P0Table2, 6, Params.ab_multiplier_per_frq.z, saturated.z);
   
+#if MESA_BUGGY_BOOL_CMP
```
christiankerl: change this to #ifdef and then simply prepend #define MESA_BUGGY_BOOL_CMP to every source file?

examples/protonect/src/opengl_depth_packet_processor.cpp
```diff
@@ -117,31 +124,81 @@ bool loadBufferFromFile(const std::string& filename, unsigned char *buffer, size
...
+      }
+    }
+    mesa_checked = true;
+    return is_buggy_mesa;
+  }
+
+  void patchMesaBug(std::string &src)
   {
+    LOG_WARNING << "Working around buggy boolean instructions in your Mesa driver. Mesa DRI 10.3+ is recommended.";
+    std::string a("MESA_BUGGY_BOOL_CMP");
```
christiankerl: prepend "#define MESA_BUGGY_BOOL_CMP\n" to src (see also my other comment)

examples/protonect/CMakeLists.txt
```diff
@@ -166,6 +166,28 @@ IF(ENABLE_OPENCL)
     LIST(APPEND RESOURCES
       src/opencl_depth_packet_processor.cl
     )
+
+    INCLUDE(CheckCSourceRuns)
```
christiankerl: move this part to a separate cmake file in cmake_modules similar to SetupLibfreenect2Threading.cmake?! keeps the main CMakeLists.txt cleaner?

xlz: It's OK. The problem is ocl-icd may be entirely different for binary package builders and users. This is only compile-time detection for the ocl-icd bug.

I think it may be fine to disable exception for all users, like [chromium](https://www.chromium.org/chromium-os/build/c-exception-support). Disabling exceptions has some benefits and seems harmless.

examples/protonect/src/shader/stage1.fs
xlz: Actually I implemented this idea first. The first issue was `#version` must be the first line of each shader file. I was not sure if the version line is intended to stay in the shader source so this. I can do it again though.

The second thing is, there are a bunch of constant parameters that can actually be reused as macros, both in shader files and in the opencl kernel, and in cuda kernel -  create a `depth_parameters.h.in` with `MEGA_BUGGY_BOOL_CMP` and embedded it as a resource, and add the rest of the parameters to it later.

examples/protonect/src/opengl_depth_packet_processor.cpp
RyanGordon: Ah okay, thanks for the explanation

xlz:
v2:

* `ocl-icd` bug detection can only happen in run-time, but the workaround requires compile-time change. So all Linux builds now disable exceptions. `ocl-icd` bug detection just prints a warning.
* Prepend `#define MESA_BUGGY_BOOL_CMP` if the Mesa bug is detected.
* `checkMesaVersion()` is replaced by prepending definitions.

xlz:
v3:

* Moved Mesa bug detection out from ShaderProgram to runtime methods, because glGetString will return NULL in the constructor.

xlz:
Rebased to master.

floe:
Could you please rebase once more, so it only shows the new commits?

floe:
Tested on Ubuntu 14.04 with `Intel Corporation 4th Gen Core Processor Integrated Graphics Controller (rev 06)`, works like a charm (and actually 10 Hz faster than OpenCL). Kudos.

xlz:
Rebased to master.

floe:
No, this was with the viewer enabled. Nevertheless, since the viewer and the OpenGL DPP are not linked together, I would assume that download and re-upload to GPU memory still have to happen?

xlz:
Yes. OpenGL should be slower because it does extra flipping on CPU.

floe:
As usual: if we don't merge it, we'll never know if it has issues. Merging.

PR #372, opened 2015-08-27
stdlib threading is only available for c++11

fran6co:
Mac OSX doesn't support thread_local, but libfreneect is not using it


floe:
Did you happen to test this on a non-Mac OS, just to be safe?

fran6co:
No, I don't have another platform to test it.

christiankerl:
threadlocal was only required for the old glew multi context opengl bindings. so it should be save to remove

Issue #144, opened 2015-01-28
Depth measurement offset


wiedemeyer:
While calibrating the Kinect2 to our robots, I saw that the depth values do not match the robots model.

I investigated this and wrote a program that compared the measured distances to the calculated distances from a chess board I used for the calibration. From each image I used the region of the board and compared each pixels (valid) depth measurement to the calculated distance by ray intersection to the board plane. Before doing this I undistorted the images.

It seems like the Kinect2 has a static offset of around 24 mm. The depth measurement is always 24 mm further away than it should be. Here are some plots of the results. This offset seems to be not related to image coordinates or distance. The error grows towards the image corners, but there the distortion is much stronger and the values get more noisy.

![plot.png](http://ai.uni-bremen.de/wiki/_media/software/plot.png)
![plot_x.png](http://ai.uni-bremen.de/wiki/_media/software/plot_x.png)
![plot_y.png](http://ai.uni-bremen.de/wiki/_media/software/plot_y.png)
![plot_xy.png](http://ai.uni-bremen.de/wiki/_media/software/plot_xy.png)

I only used the OpenCL depth packet processor. So I can't tell right now, if it is just related to that specific implementation, all implementation or the device itself. Tomorrow I will try out the OpenGL and CPU depth packet processor as well. Did someone experienced a similar behavior until now?

@christiankerl: Do you think it could be a problem of the packet processor implementation?

I added the code for the depth calibration to the [calibration tool](https://github.com/code-iai/iai_kinect2/tree/devel/kinect2_calibration) of my [ROS Kinect2 bridge](https://github.com/code-iai/iai_kinect2/tree/devel/). If the depth values are decreased by the calculated offset the resulting depth cloud perfectly matches the model of our robots.

christiankerl:
I guess the problem is that the xtable and ztable are used in the depth calculations, but they are only valid for the device I used during the memory dump. Actually they should be computed from some calibration values stored on each device. However, the formula is unknown

sh0:
I heard that Kinect 2 has slight distance offset depending on what material you are observing. I'm not a specialist on ToF cameras, but I guess that when the material is highly reflective then the reflected light intensity is stronger and will activate the photosensitive element faster. Reportedly there can be 1-2 cm difference when comparing specular metal to a material with very diffuse reflection. I haven't tested for this effect myself so I really can't say how true all of this actually is.

wiedemeyer:
[...]

PR #379, opened 2015-09-03
Generate depth tables with camera parameters

xlz:
The xtable, ztable, and 11to16 LUT can now be generated with camera parameters at runtime according to analysis in #144. The tables are generated during `Freenect2Device::start()`, and passed to depth processors.

Users can provide custom camera parameters at runtime with new API: `setIrCameraParams()`, and `setColorCameraParams()`, and depth processors will use those instead of USB queried parameters.

File loading functions in depth processors are removed.

Hardcoded table binary files are removed.


floe:
Excellent work, could you perhaps add a short comment explaining the functionality of xtable/ztable itself (since you've reverse-engineered it now fully)?

xlz:
v2:

* Added comments about 8192 and 6250.0/3 magic numbers
* Added comments about the nature of x/ztable
* Consolidated X/Z/LUTables generation functions into `struct IrCameraTables`.

Old comments can be found here: https://github.com/xlz/libfreenect2/commit/866fdb2dbab42cd7727db3dd52a8260cd964be21

xlz:
Rebased to master.

xlz:
Rebased to master.

floe:
Just tested this, works as expected. Purely from visual inspection, I can't tell any difference between the two versions; I think @wiedemeyer had done some measurements?

xlz:
To tell a difference, extract two frames from before and after, `imagesc(abs(before-after))` will visualize the difference.

To absolutely prove the correctness, another set of x/z/lut tables have to be extracted and compared with tables computed here. Or another round of calibration set with `setIrCameraParams()` with external intrinsic parameters.

xlz:
You might see some patterns obtained in http://designfankhauser.ch/files/fankhauser_2014_kinect_v2.pdf fig 5.

It is possible some of the errors analyzed in the paper come from the discrepancy of the intrinsic parameters used for image warping (external) and depth decoding (hardcoded).

xlz:
diff
This is the absolute error between before and after this patch (pixel values are the average of 100s image, only showing pixels with standard deviation less than 4). 

Histogram of the error between before and after. 
hist

floe:
Ah, this is very interesting - AFAICT, the error distribution means that there wasn't a systematic error before, just a different error distribution?

xlz:
The above plots might not be meaningful. Probably most of it is just noise and not showing the bias.

Here are another two plots showing the theoretical errors in the final depth values between using the hardcoded parameters, and (1) the embedded intrinsic parameters of my Kinect; (2) iai_kinect2's sample intrinsic parameters:

diff1
diff2

I guess this really depends on the particular calibration parameters. At the edge the error is as much as a few millimeters, but another problem is data at the edge has poor reflection and poor reliability so is probably discarded for reliable computation. In the middle the error should be tiny.

floe:
We should get this tested a little more widely, so I'm merging it now.

Issue #392, opened 2015-09-16
[Question] From registered libfreenect2::Frame to pcl::Pointcloud<pcl::PointXYZRGBA>


Tabjones:
Hello everyone, first of all thanks for the amazing work.

I'm looking at Protonect.cpp source code and i see that after ` registration->apply(rgb, depth, &undistorted, &registered);` (line 203) I should have a color-depth registered image in `registered` Frame. In fact that same frame gets sent to the viewer for the user to see.

Now what i want to do is convert that frame into a PointCloud class (from pcl library), as far as i know it should contain all the necessary information to construct a point cloud: it has width, height and data.

My question is, how can I access the data pointer to extract euclidean coordinates (x,y,z) and the color associated with each point ?
Are there already some helper functions doing this that i'm missing ?

xlz:
For `xi=0..511, yi=0..423`, normalized distorted coords `(xd, yd) = ((xi+0.5-cx)/fx, (yi+0.5-cy)/fy)`.
Normalized undistorted coords `(xu, yu)` are obtained from `kinect2_undistort()`.
`(X, Y, Z) = (xu*Z, yu*Z, Z=undistorted[yi*512+xi]), (R, G, B)=registered[yi*512+xi]`.

The undistort function `kinect2_undistort()` was reverse engineered from the official Kinect SDK. See here https://github.com/OpenKinect/libfreenect2/issues/41#issuecomment-72022111

But that function was subsumed by a reverse pixel mapping method in the `Registration` class.

Registration is pretty high level for the library. No helper at the moment.

floe:
@xlz, the second part of your answer is a better explanation IMHO; main reasoning for having undistorted images is that a) they are better suited to purely image-based processing, e.g. depth filtering, and b) they are a by-product of color-to-depth mapping anyway. In particular, the depth-mapped color image is also implicitly undistorted/rectified, so if you want the corresponding depth values for a color pixel from there, you need to use `undistorted`.

@Tabjones This shouldn't actually be very complex to implement, feel free to submit a pull request with a suitable helper function (although we probably won't add a dependency on PCL just for that). Maybe it can be implemented as a header-only template, to be included alongside PCL?

Tabjones:
@xlz ok it think i get it. Only thing it's not yet clear to me: what are cx,cy and fx,fy ? Some kind of distortion parameters from intrinsic calibration? I should be able to access those from camera parameters?

[...]

PR #397, opened 2015-09-20
First prototype of computeCoordinates of point cloud, yet to be tested

Tabjones:
I started the PR mentioned  in #392 so you can see the changes, i'll test the function tomorrow and let you know. You can wait to merge until some tests are performed.
Cheers

### DIFF
--- a/examples/protonect/include/libfreenect2/registration.h
+++ b/examples/protonect/include/libfreenect2/registration.h
@@ -29,10 +29,6 @@
 #ifndef REGISTRATION_H_
 #define REGISTRATION_H_
 
-#ifdef LIBFREENECT2_WITH_CX11_SUPPORT
-#include <array>
-#endif
-
 #include <string>
 #include <libfreenect2/config.h>
 #include <libfreenect2/libfreenect2.hpp>
@@ -40,10 +36,6 @@
 
 namespace libfreenect2
 {
-
-#ifdef LIBFREENECT2_WITH_CX11_SUPPORT
-typedef std::array<float, 512*424> Array;
-#endif
 
 /** Combine frames of depth and color camera. */
 class LIBFREENECT2_API Registration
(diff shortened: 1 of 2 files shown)


xlz:
I don't think you need to use C++11 features for this.

Tabjones:
what you mean ? the use of STL array is overkill ?!

xlz:
> the use of STL array is overkill

Yes

Tabjones:
Ok, you are probably right, i'll remove std::array.
Those pointers will leak yes and also the idx is wrong as you said, will probably cause segfault.
Didn't have time to test/fix those, but i noticed them.

Also i was thinking that probably it is more efficient to create a function that computes coordinates of a single point, given the indexs. something like this:
`void computeCoordinate (Frame* undistorted, Frame* registered, int row, int col, float& x, float& y, float& z, float& rgb)` So that the user can fill the point cloud with a single loop. Returning an array would force the user to reloop them all in his program to fill PointXYZRGB data.
Also more user-safe in terms of memory leaks, cause the user would not have to use float*.

If you agree i'll modify it accordingly

xlz:
Starting with a basic one should be good.

```cpp
void getPointXYZRGB(Frame* undistorted, Frame* registered, int row, int col, float& x, float& y, float& z, unsigned int& rgb)
```

Tabjones:
Tried it and it looks acceptable. 
screenshot from 2015-09-22 10 44 15

Had to convert rgb into float, so that it can be directly assigned to pcl::PointXYZRGB::rgb.
There's lot of noise around borders and color-depth registration doesnt look precise, but i'm satisfied for now.

floe:
Noise around borders is a known artifact of TOF cameras AFAIK. The registration might also get a bit better once #379 has been merged.

moonsh:
Firstly thank your hard work.
I would like to approach X,Y,Z value so I coded like this and 

	for (int xi = 0; xi < 512; xi++)
	{
		for (int yi = 0; yi < 424; yi++)

		{
	registration->getPointXYZRGB(&undistorted, &registered, xi, yi, X, Y, Z, RGB);
	
		}
	}

Tabjones:
i think you inverted rows and columns: i mean what you called xi should iterate through rows, i.e. the height (424). Something like this:
```
  for (int r=0; r<424; ++r)
    for (int c=0; c<512; ++c)
      registration->getPointXYZRGB(undistorted, registered, r,c, x,y,z,rgb);
```

moonsh:
Thanks . Here I have one more question. Image described by points is not dense and even rough.  Is it maximum resolution kinect v2 can do ? or is there anyway to Increase resolution ?

Tabjones:
Yes it is not dense, cause it could contain NaN values.
About resolution, I believe that is the sensor provided resolution for depth (512x424), because that is the actual resolution of the IR camera.
In principle you could increase resolution by performing upsampling and/or other smoothing filters of the depth image, but of course you'll have to reapply a modified registration of color-depth again.

If that is what you need there's another package (that uses libfreenect2) that does exactly this upsampling, giving you point clouds at quarter-hd or full hd (1920x1200). https://github.com/code-iai/iai_kinect2, definitely worth a try.
Cheers

moonsh:
Ah I see !  Sorry for annoying you again . How about RGB values ? For visualization, R,G and B values are needed separately but here I can get only rgb value .

Tabjones:
umh, not sure but you could try unpacking as described here http://docs.pointclouds.org/1.7.0/structpcl_1_1_point_x_y_z_r_g_b.html

moonsh:
Here is result I got  but some problem is exist . As you can see some points are missing in above figure. What factors makes this missing points? too near or too closer ? or  Am I taking point data too quick ?

xlz:
@moonsh Maybe reflection is too strong and exceeds the sensing limit. Some of the low quality data is also removed from output.

I have also seen flat surface with patches without data. I wonder if there is any way to get around this.

PR #402, opened 2015-09-23
Get rid of CL_DEVICE_TYPE_CUSTOM

floe:
Caused several issues with older OpenCL SDKs, can be removed safely.


xlz:
this is ok.

Opencl needs two other fixes. First, get cmake's latest findopencl.cmake instead of ours. Second, include cl.hpp, so no more mac osx hack is needed, and also Debian's opencl-headers no longer has cl.hpp.

These two have liberal licenses and no licensing problem, but perhaps you want a separate file to list their licenses for the ease of redistribution.

floe:
FindOpenCL.cmake is part of cmake 3.x, but we require only cmake 2.8.something, so it might not be backwards-compatible?

xlz:
It looks like all modules used in FindOpenCL.cmake (CheckSymbolExists, CMakePushCheckState, FindPackageHandleStandardArgs) are available as early as 2.8.12.

PR #376, opened 2015-09-01
0.1 release build system restructuring

xlz:
This implements the source code and build system restructuring in #314 (rename examples/protonect to src), which is intended to the last major PR before 0.1 release. All issues for 0.1 (#233) except x/ztable hardcoding are resolved after this PR.

This PR is rebased on #368 (OpenGL fixes) and #364 (logging system). Commits from #368 and #364 are shown here, but **this PR starts from `Use BUILD_SHARED_LIBS to control library type`**. Commits from other PRs can be commented in their respective PRs.

* `Use BUILD_SHARED_LIBS to control library type` implements static/shared build configuration, requested in #292 and #263, but reverting #276.
* `Allow custom RPATH settings` deprecates libusb embedding #283. Package distributors are now able to specify the custom libusb path to use for libfreenect2.
* `Raise CMake version requirement to 2.18.12.1` fixes a bug #363
* The rest of the commits are about restructuring. They all have distinct descriptions.

There are some ABI changes. I try to minimize the exposed API before the release for forward compatibility. In the commit 2f186a5894988e14f1700f0a70d0e8b55d53a9bf "Separate public and internal API", several headers are moved to an internal header directory and are no longer installed (no change for internal code though).  Protonect in the examples directory can be built with installed libfreenect2: `cd examples; cmake .. -Dfreenect2_DIR=/tmp/libfreenect2/lib/cmake/freenect2; make; ./Protonect`. This will validate the API usage by third-party. And in the commit "Use CMake to generate `LIBFREENECT2_API` macro", symbols become hidden by default, except those explicitly marked with `LIBFREENECT2_API`. Inspecting libfreenect2.so with `nm -g --defined-only -C libfreenect2.so` should reveal only the necessary ones are exported.

This has been tested on Ubuntu/Intel.

I will rebase the PR with updates in the repo.

PR #403, opened 2015-09-23
fixed tinythread include

giacomodabisias:
Fixed this include path


xlz:
What problem does this solve?

giacomodabisias:
tinythread.h not found when compiling using libfreenect2

xlz:
tinythread.h will be removed from public API. So this is not needed.

floe:
I think it will be obsoleted by #376, which I will merge now.

PR #423, opened 2015-10-13
Add teapot configuration file

ioquatix:
This is for building this library with teapot (http://teapot.nz). It may need to be updated from time to time.

### DIFF
--- a/teapot.rb
+++ b/teapot.rb
@@ -1,0 +1,29 @@
+
+#
+#  This file is part of the "Teapot" project, and is released under the MIT license.
+#
+
+teapot_version "1.0.0"
+
+define_target "libfreenect2" do |target|
+	target.build do
+		source_files = target.package.path
+		cache_prefix = Files::Directory.join(environment[:build_prefix], "libfreenect2-#{environment.checksum}")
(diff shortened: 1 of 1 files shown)


floe:
Didn't know about teapot before; how does it deal with installing prerequisites (libusb etc.)?

ioquatix:
So, this is an initial package file to get the ball rolling.

Ideally, teapot builds are idempotent, so it would explicitly specify all dependencies and they could be locked to specific versions. Right now, teapot is just invoking `cmake` directly so all dependencies must be managed by the host OS package manager.

ioquatix:
I'm currently throwing around the ideas of having teapot directly interact with CMake and also have a semi-official centralised repository of packages. I don't like reinventing the wheel though, but teapot does require packages to supply build rules unlike, say, CMake which has them built in, or Make which doesn't really. Because this requirement of packages to supply build rules which in general are usually common for many scenarios it might make sense to share them in an organised way.

Adding dependencies is really easy but if you want to go down that route I'd suggest using teapot to build the entire project and avoid the dependency on CMake for the teapot build. If you'd like to explore this option let me know.

ioquatix:
I'm no longer interested in this and it's out of date.

Issue #390, opened 2015-09-15
libusb-1.0.20 release and PPA


floe:
As pointed out by @xlz, libusb has a new release 1.0.20. Luckily, this now does work with libfreenect2, at least on Intel controllers.

I've been so fed up with the whole libusb-multiple-version-RPATH mess that I quickly created an Ubuntu PPA for the new version: https://launchpad.net/~floe/+archive/ubuntu/libusb

Not sure how to handle this for Debian, though - suggestions?

xlz:
I submitted a bug report to Debian for the new release. https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=799069

Hopefully the new release is packaged within one or two months, then Debian unstable, testing will have it. Stable probably will not have it. In the mean time, the build-from-source instructions are still useful. Once a binary package is released, users can download that.

xlz:
Debian has updated their release. Very fast.

floe:
OK, to summarize how to get libusb-1.0.20 ATM:
* Ubuntu: use my PPA on 14.x, or download .deb files from Launchpad
* Debian: use unstable/testing, or download .deb files manually
* MacOSX: install via `brew` (?)
* Windows: get binaries from libusb homepage

xlz:
Windows: still have to build from source with joshblake branch. libusb 1.0.20 does not seem to include Windows usb3 iso support.

JoshBlake:
> libusb 1.0.20 does not seem to include Windows usb3 iso support.

(╯°□°）╯︵ ┻━┻

metropt:
How do I check if I have the newest version installed?

xlz:
@metropt `apt-add-repository ppa:floe/libusb && apt-get update && apt-get install libusb-1.0-0-dev`

floe:
Actually installed version can be shown with `dpkg -l libusb-1.0*`. If you have 1.0.20 there, then you can also delete the `libfreenect2/depends/libusb*` folders, maybe it's still picking up the previous self-compiled version.

floe:
I should add that right now, this repo only publishes packages for 14.04 (trusty). I'll change that soon.

ahundt:
can this update also be verified to make its way in on OS X's homebrew?

ahundt:
looks like someone submitted a pull request over there but there is a small bug that would have to be resolved before it is merged https://github.com/Homebrew/homebrew/pull/44038

fran6co:
[...]

PR #435, opened 2015-10-20
Fixes missing subpackets in OS X

fran6co:
Had some issues in OS X with libusb missing some subpackets for no good reason and some other weird errors as noted in => https://github.com/OpenKinect/libfreenect2/issues/390

Looking at libfreenect 1.x code I found that https://github.com/OpenKinect/libfreenect/blob/master/src/usb_libusb10.h#L33 uses different parameters for OS X, just ported that from there.

### DIFF
--- a/src/libfreenect2.cpp
+++ b/src/libfreenect2.cpp
@@ -43,6 +43,14 @@
 #include <libfreenect2/protocol/command_transaction.h>
 #include <libfreenect2/logging.h>
 
+#ifdef __APPLE__
+  #define PKTS_PER_XFER 128
+  #define NUM_XFERS 4
+#else
+  #define PKTS_PER_XFER 8
+  #define NUM_XFERS 60
+#endif
+
 namespace libfreenect2
 {
 using namespace libfreenect2;
@@ -464,8 +472,8 @@ bool Freenect2DeviceImpl::open()
     return false;
   }
 
-  rgb_transfer_pool_.allocate(50, 0x4000);
-  ir_transfer_pool_.allocate(80, 8, max_iso_packet_size);
+  rgb_transfer_pool_.allocate(20, 0x4000);
+  ir_transfer_pool_.allocate(NUM_XFERS, PKTS_PER_XFER, max_iso_packet_size);
 
   state_ = Open;
 
@@ -585,7 +593,7 @@ void Freenect2DeviceImpl::start()
 
   LOG_INFO << "submitting usb transfers...";
   rgb_transfer_pool_.submit(20);
-  ir_transfer_pool_.submit(60);
+  ir_transfer_pool_.submit(NUM_XFERS);
 
   state_ = Streaming;
   LOG_INFO << "started";


xlz:
Overall if this improves packet loss on Mac OSX, I'd say it's good to merge.

Though changing the size of transfer pool is risky if you don't understand why this change works. I thought about doing this long for several previous bugs where errors were reported for not enough bandwidth.

My basic understanding of is transfer pool will reserve USB iso bandwidth with guaranteed latency w.r.t the size of the transfer pool. Reserving too much will cause not enough bandwidth error. Reserving too little will risk having lots of packet loss. And there is certain abstraction from usbfs buffer to "transfer" (Linux) so the number of transfers is not simply the bandwidth reserved.

Hopefully this can be explained with USB 3.0 spec and libusb code later.

fran6co:
That's why I changed it only for OS X, no idea how it works for the rest of the platforms but it works fine on mine and libfreenect 1 works fine too so I figured we could use the same config. I tried to keep PKTS_PER_XFER* NUM_XFERS around the same, it's 128*4=512 for OS x and 60*8= 480 for the rest.

xlz:
This uses 128 packets per transfer. I don't know your `max_iso_packet_size` (in libfreenect2.cpp). Assuming 0x8400, that is 4224KB per transfer. A single depth image requires raw data of 10x352x424 bytes, 1457.5 KB. So a single transfer contains data for 2 depth images, and incomplete data for the 3rd depth image, which would have to wait for the next transfer.

I think here unnecessary and non-uniform latency is introduced.

xlz:
kIOReturnIsoTooOld comes from this pseudo stack trace:

* ReadIsochPipeAsync() [libusb/os/darwin_usb.c:1587](https://github.com/libusb/libusb/blob/960a6e7506c4d0d59052056ae2ebfffbe8587175/libusb/os/darwin_usb.c#L1587)
* IOUSBInterfaceClass::ReadIsochPipeAsync(), IOUSBFamily...
* IOConnectCallAsyncScalarMethod(..., kUSBInterfaceUserClientReadIsochPipe, ...)
* IOUSBInterfaceUserClientV2::_ReadIsochPipe
* DoIsochPipeAsync()
* IOUSBPipe::Read(???)
* IsocIO()
* DoIsocTransfer()
* IsocTransaction()
* AppleUSBXHCI::UIMCreateIsochTransfer() - no public source code, use AppleUSBEHCI as reference

UIMCreateIsochTransfer() is the only place that generates kIOReturnIsoTooOld. kIOReturnIsoTooOld is returned immediately for the async read when `command->GetStartFrame() < pEP->firstAvailableFrame` or `command->GetStartFrame() < (curFrameNumber - maxOffset)`.

Not sure what this means yet. 

There are some more things you can try. I see a lot of `USBLog()` lines which kprintf to "kernel log" (is this a thing on Mac OSX?) Probably you can dig that out with https://developer.apple.com/hardwaredrivers/ "Debug releases of the IOUSBFamily kernel extension".

The second thing is this kIOReturnIsoTooOld seems to be generated by some static conditions. Maybe explore how `frame` changes in [libusb/os/darwin_usb.c:1587](https://github.com/libusb/libusb/blob/960a6e7506c4d0d59052056ae2ebfffbe8587175/libusb/os/darwin_usb.c#L1587)
```
    kresult = (*(cInterface->interface))->ReadIsochPipeAsync(
          cInterface->interface, pipeRef, transfer->buffer, frame,
          transfer->num_iso_packets, tpriv->isoc_framelist, darwin_async_io_callback,
          itransfer);
```
and maybe certain conditions really make `frame` here outdated somehow.

fran6co:
I'm increasing it to 136, that should cover it.
In any case I'm going to setup a machine to run the kinect2 all day and see if I get the kIOReturnIsoTooOld error or the missing packets. The master version throws them from the go, so this is an improvement already.

fran6co:
It's been around 2 hours and no errors so far, it looks like this fixed the issue.

xlz:
I think you should decrease it to reduce latency.

fran6co:
I don't understand what's the problem with the latency... I mean, the only problem is the time it takes to retrieve the extra data for the next frame, I don't think that extra time is that big compared to all the processing it's doing. In any case I'll reduce it to fix a single depth image in the transmission.

xlz:
The application is realtime and latency sensitive. If you batch two depth images in one libusb transfer, the two images are processed together, resulting in non-uniform performance down the pipeline. And by the time the second image is processed, the first image is already outdated.

fran6co:
Ok, I reduced it to 64 that's around 1 depth image.

xlz:
I found the source code of AppleUSBXHCI https://opensource.apple.com/tarballs/IOUSBFamily/IOUSBFamily-560.4.2.tar.gz

xlz:
Can you try libusb with this change:
```
  /* schedule for a frame a little in the future */
  frame += 64;
```
`frame += 64` was originally added in libfreenect ("+= 2 was too short"):
https://github.com/OpenKinect/libfreenect/commit/562a7b28eb7ab7d7f361d1d5f3e4396b4ffc9dca

I looked at AppleUSBXHCI, and `frame` should be scheduled a little bit a head of `pEP->firstAvailableFrameNumber`, otherwise other transfers could race beyond that. (But I don't see any race condition though). If `4` is not enough, try 64.

fran6co:
I'll check it out, in any case keep it mind that the errors are totally gone with this PR using homebrewed libusb 1.0.20.

fran6co:
Ok, just tested. I'm not getting the kIOReturnIsoTooOld errors anymore but I still get all the rest.

xlz:
OK, let's go with this for Mac users for now with increased latency. It's still a mystery why 60 concurrent transfers are giving MacOSX a hard time though.

xlz:
One more note,  when receiving two depth images together the depth processor will not be ready for the next one, resulting in "skipping depth packet" for every other depth image. I tested 128 packets per transfer on Linux and got that.

fran6co:
Uhm, I think the only way to solve that is to add some buffering in between. Any other ideas?

xlz:
I don't see any problem for MacOSX if your tests say OK. Just documenting relevant information about libusb behavior.

xlz:
Sorry, there was an error in my calculation.

It's 352x424 shorts, not 352x424 bytes. Therefore for each depth images, there are `10*(352*2*424+38*4) = 298648` bytes transferred. Assuming packet size of 0x8400, that is 89 packets at minimum for each depth image. There are idle room between each "subpacket", so 128 packets per transfer is actually probably the correct number if you want to fit one depth image in one transfer. Sorry for the confusion.

xlz:
I am looking at if using less transfers and more packets per transfer is a good idea for Linux. One problem with Linux kernel is that kmalloc of more than 2MB is discouraged, so transfers with buffer size of more than 2MB probably are not a good idea. MacOSX's kernel may or may not have the same issue.

The transfer timing pattern looks like this (`a`: a 33792-byte packet; `b`: a 28312-byte packet; `.`: an empty packet; each `aaaaaaaab` is a 298648-byte depth "sub-layer")
```
aaaaaaaab.aaaaaaaab...aaaaaaaab.aaaaaaaab...aaaaaaaab.aaaaaaaab..aaaaaaaab..aaaaaaaab...aaaaaaaab................................aaaaaaaab.................................................................................................................................
aaaaaaaab.aaaaaaaab..aaaaaaaab..aaaaaaaab..aaaaaaaab..aaaaaaaab..aaaaaaaab.aaaaaaaab...aaaaaaaab................................aaaaaaaab.................................................................................................................................
```
The period of such transfers is 266 or 267 packets. Each packet takes a service interval of 125us to transmit. Therefore each period takes `266.6 * 125us = 33.33375ms`, which is exactly the Kinect 2 frame rate 30Hz.

The 10-th "sub-layer" is not actually used for depth computation, so the first 9 "sub-layers" takes less than 100 packets to transmit. After the first 9 sub-layers, the rest of the period does not have any actual data that requires an actual buffer to take. Perhaps we can get away with not submitting a transfer at all and wait until the next period begins and then submit another transfer.

So maybe in the future, a stream "non-parser" will just take a libusb buffer, check its lengths, and send that to depth processors without doing any copying based parsing. This will also reduce the number of transfers required.

fran6co:
Back to 128. I'm getting skipped packets warnings but not more than before.

xlz:
There is no getting away with not submitting transfers when idle.
```
 * Device drivers must explicitly request that repetition, by ensuring that
 * some URB is always on the endpoint's queue (except possibly for short
 * periods during completion callbacks).  When there is no longer an urb
 * queued, the endpoint's bandwidth reservation is canceled.
```

victorcoder:
I think I'm suffering this problem in OSX 10.11 but with libfreenect, The device just stop working with glview.

Tested the Kinect 360 with another Mac with 10.10 and it works well so I thinks it's related to the libusb 1.0.20

Relevant logs https://gist.github.com/victorcoder/cb259ce2eb3b04682134

PR #441, opened 2015-10-22
makes the map for storing the color offset for each depth pixel a function parameter in order to make the user decide the allocation policy

giacomodabisias:
makes the map for storing the color offset for each depth pixel a function parameter in order to make the user decide the allocation policy

### DIFF
--- a/src/registration.cpp
+++ b/src/registration.cpp
@@ -100,9 +100,10 @@ void Registration::apply( int dx, int dy, float dz, float& cx, float &cy) const
  * @param [out] registered Image color image for the depth data (512x424).
  * @param enable_filter Use a depth buffer to remove pixels which are not visible to both cameras.
  * @param [out] bigdepth If not \c NULL, mapping of depth onto colors (1920x1082 'float' frame).
+ * @param [out] bigdepth If not \c NULL, map for storing the color offset for each depth pixel.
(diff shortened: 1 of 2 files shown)


src/registration.cpp
```diff
@@ -100,9 +100,10 @@ void Registration::apply( int dx, int dy, float dz, float& cx, float &cy) const
  * @param [out] registered Image color image for the depth data (512x424).
  * @param enable_filter Use a depth buffer to remove pixels which are not visible to both cameras.
  * @param [out] bigdepth If not \c NULL, mapping of depth onto colors (1920x1082 'float' frame).
+ * @param [out] bigdepth If not \c NULL, map for storing the color offset for each depth pixel.
```
xlz: bigdepth -> color_depth_map. And its required size?

floe:
Since this doesn't really change the interface or the internal operation and just provides an option which some people may find useful, I'll include this for 0.2.

giacomodabisias:
Thanks

xlz:
This breaks the ABI. If this is to be merged, it is before 0.1.

floe:
OK, not a problem. Fine for 0.1, too.

ludiquechile:
Hi I think there's an issue on this commit, it appears that the method apply() of class RegistrationImpl is missing the parameter color_depth_map.

Issue #156, opened 2015-02-06
RGB camera internals (was: question about Auto Gain Control of Kinect 2 driver.)


enddl22:
Hello, forks.
I would like to ask about Auto Gain Control of Kinect 2 driver. Are there any ways of controlling AGC or AEC of Kinect2? I want to turn off these functions since they change illumination that makes life more challenge. It would be great if anyone has tried this before or has any suggestions for it.

floe:
As far as I know, we don't have any background info on the internals of the RGB camera (i.e. commands, registers etc.). Does the official SDK export any related functions? If it doesn't, then it will be very difficult to determine how to access these functions via libfreenect2.

JoshBlake:
The official SDK does not support any type of configuration of the sensor,
although this is more a design decision than a hardware capability. The SDK
contains a service that allows multiple apps to connect, and each apps
shares the sensor without owning it. Thus, the SDK does not allow one app
to configure anything that would change the hardware or the expectations
from another app.

stfuchs:
Apparently the official SDK seems to offer some functions to retrieve the RGB camera settings:
https://msdn.microsoft.com/en-us/library/microsoft.kinect.kinect.icolorcamerasettings.aspx
Is there a way to get these for libfreenect2 as well?

floe:
Not at the moment - we'd need to know what USB command is sent by the official SDK to retrieve these settings.

stfuchs:
Do you guys have any documentation how you figured out the other USB commands? Or do you have any general reference? Maybe I'll find some time to play around. Btw, great work so far :)

floe:
There are two possible ways:
* if you're running Windows 8/10 natively, you can use MMA to capture USB traffic: https://msdn.microsoft.com/en-us/library/windows/hardware/dn741264%28v=vs.85%29.aspx
* if you're running Windows in a VM (e.g. VirtualBox), route the Kinect USB connection to the VM and then use Wireshark on the host to capture USB traffic: https://wiki.wireshark.org/CaptureSetup/USB

christiankerl:
I think the information is in the rgbpacketfooter seehttps://github.com/OpenKinect/libfreenect2/blob/master/examples/protonect/src/rgb_packet_stream_parser.cpp you just need to extend the api to pass this information to the frame listened...

floe:
Ah, so maybe `unknown3` is the gamma value?

stfuchs:
[...]

PR #465, opened 2015-11-10
Feature/camera settings

stfuchs:
### make the color camera settings available in libfreenect2::Frame.
see #156

### DIFF
--- a/include/libfreenect2/frame_listener.hpp
+++ b/include/libfreenect2/frame_listener.hpp
@@ -62,7 +62,7 @@ class LIBFREENECT2_API Frame
     width(width),
     height(height),
     bytes_per_pixel(bytes_per_pixel),
-    exposure(1.f/30.f),
+    exposure(30.f),
     gain(1.f),
     gamma(1.f)
   {
--- a/include/libfreenect2/frame_listener.hpp
+++ b/include/libfreenect2/frame_listener.hpp
@@ -54,11 +54,17 @@ class LIBFREENECT2_API Frame
   size_t height;          ///< Number of lines in the frame.
   size_t bytes_per_pixel; ///< Number of bytes in a pixel.
   unsigned char* data;    ///< Data of the frame (aligned).
+  float exposure;
+  float gain;
+  float gamma;
 
   Frame(size_t width, size_t height, size_t bytes_per_pixel) :
     width(width),
     height(height),
-    bytes_per_pixel(bytes_per_pixel)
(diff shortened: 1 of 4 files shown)


include/libfreenect2/frame_listener.hpp
```diff
@@ -54,11 +54,17 @@ class LIBFREENECT2_API Frame
...
+  float exposure;         ///< Get the exposure time set by the color camera.
+  float gain;             ///< Get the gain set by the color camera.
+  float gamma;            ///< Get the gamma level set by the color camera.
 
   Frame(size_t width, size_t height, size_t bytes_per_pixel) :
     width(width),
     height(height),
-    bytes_per_pixel(bytes_per_pixel)
+    bytes_per_pixel(bytes_per_pixel),
+    exposure(30.f),
```
floe: If this is in milliseconds, should the default value then be 33.3f?

stfuchs: To be honest, I have no idea what might be a good default value. It ranges from < 0.1 up to 66.4719085693 when it's fully covered. 

So you mean `1.f/30.f * 1000.f`?

floe: Yes, something along those lines. 66.47 in milliseconds would be 15 Hz, and I remember something that the color camera does indeed drop the framerate to 15 Hz in low-light situations. So default at 30 Hz would then be ~ 33.3.

xlz: It should be 0.0f or NaN? There is no gamma for IR or depth images.

stfuchs: Agree. But 1. would be the identity transfer function for gamma and should also work fine as default right?

xlz: Basically give it in invalid value because it will be filled in if working correctly.

stfuchs: Ok. NaN would require `#include <limits>`, is that fine with you?

xlz: If you ask I'd go with 0.0f.

floe:
I'll merge this for 0.1, as the API remains unchanged (even though it will break the ABI AFAICT).

xlz:
Can this be merged? The documentation PR is after this.

floe:
Ah, sure, had sort of forgotten about that one, sorry.

Issue #205, opened 2015-04-11
Abort trap: 6 on OSX Yosemite 10.10.3


omarev-git:
Hello,
Im was working fine with OS X 10.9 (Mavericks) but by requirements of our office I made the upgrade to Yosemite (10.10.3) everything compile good but when I tray to run  Protonect I have the following error:
[Freenect2Impl] enumerating devices...
[Freenect2Impl] 12 usb devices connected
[Freenect2Impl] found valid Kinect v2 @21:15 with serial 500086241942
[Freenect2Impl] found 1 devices
[Freenect2DeviceImpl] opening...
[Freenect2DeviceImpl] opened
[Freenect2DeviceImpl] starting...
[Freenect2DeviceImpl] ReadData0x14 response
92 bytes of raw data
0x0000:  00 00 12 00 00 00 00 00 01 00 00 00 43 c1 1f 41    2e2e2e2e2e2e2e2e2e2e2e2e432e2e41
0x0010:  00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00    2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e2e
0x0020:  0a 21 33 55 c2 00 17 20 00 08 00 00 10 00 00 00    2e2133552e2e2e202e2e2e2e2e2e2e2e
0x0030:  00 01 00 00 00 10 00 00 00 00 80 00 01 00 00 00    2e2e2e2e2e2e2e2e2e2e802e2e2e2e2e
0x0040:  31 33 00 00 00 01 07 04 47 4d 58 36 31 30 2e 31    31332e2e2e2e2e2e474d583631302e31
0x0050:  58 00 00 00 00 00 00 00 00 00 00 00                582e2e2e2e2e2e2e2e2e2e2e

[Freenect2DeviceImpl] ReadStatus0x090000 response
4 bytes of raw data
0x0000:  5d 26 00 00                                        5d262e2e

[Freenect2DeviceImpl] ReadStatus0x090000 response
4 bytes of raw data
0x0000:  5f 26 00 00                                        5f262e2e

[Freenect2DeviceImpl] enabling usb transfer submission...
[Freenect2DeviceImpl] submitting usb transfers...
[Freenect2DeviceImpl] started
device serial: 500086241942
device firmware: 4.3.3912.0.7
[DepthPacketStreamParser::handleNewData] not all subsequences received 512
[...]

PR #467, opened 2015-11-15
Protonect Window Fix

mbgeek:
On Mac, only the lower left quadrant was being rendered with the live frames. Also there was no dynamic handling of the window resizing. 

Tested only on Mac with retail Kinect v2 in 'cl' mode. Couldn't test in gl mode having the same problem as mentioned in issue #205

### DIFF
--- a/examples/viewer.cpp
+++ b/examples/viewer.cpp
@@ -1,7 +1,10 @@
 #include "viewer.h"
 #include <cstdlib>
 
-Viewer::Viewer() : shader_folder("src/shader/")
+
+Viewer::Viewer() : shader_folder("src/shader/"), 
+                   win_width(1280), 
+                   win_height(800)
 {
     // init glfw - if already initialized nothing happens
     int init = glfwInit();
@@ -32,7 +35,7 @@ void Viewer::initialize()
 #endif
     //glfwWindowHint(GLFW_VISIBLE, debug ? GL_TRUE : GL_FALSE);
 
-    window = glfwCreateWindow(1280, 800, "Viewer (press ESC to exit)", 0, NULL);
+    window = glfwCreateWindow(win_width, win_height, "Viewer (press ESC to exit)", 0, NULL);
     if (window == NULL)
     {
         std::cerr << "Failed to create opengl window." << std::endl;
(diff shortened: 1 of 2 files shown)


xlz:
Can you explain which part of this fixes "only the lower left quadrant was being rendered with the live frames"?

examples/viewer.cpp
```diff
@@ -135,7 +150,7 @@ bool Viewer::render()
     // wipe the drawing surface clear
     glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 
-    GLint x = 0, y = 0, width = 600, height = 400;
```
mbgeek: @xlz Here's the main culprit: hardcoding the frame width and height instead of using relative sizing based on the actual window width and height.

xlz: I'm no graphics expert but I don't see a logic connection here. Do you mean if a texture does not totally fit in the viewport, then the whole texture area is not updated?

mbgeek: Essentially the viewport being rendered is set to be a hardcoded dimension versus one relative to the actual window size. The result prior to the fix is all 4 frames were rendering in just one section of the window as opposed to filling the window. Then when it was filling the window, resizing the window smaller would clip the render because it wasn't dynamically changing the viewports.

ahundt:
I can confirm I see the problem this is intending to fix

floe:
I'm currently travelling, so no way to test this - any other test reports?

RyanGordon:
This only seems to show one of the color views in the window instead of all 4 as is on master right now

xlz:
This only shows the color image. No depth or IR images are shown. Resizing works. (Debian, Intel).

examples/viewer.cpp
```diff
@@ -106,10 +109,22 @@ void Viewer::initialize()
...
 
+void Viewer::winsize_callbackstatic(GLFWwindow* window, int w, int h)
+{
+    Viewer* viewer = reinterpret_cast<Viewer*>(glfwGetWindowUserPointer(window));
+    viewer->winsize_callback(window, w, h);
+}
+
+void Viewer::winsize_callback(GLFWwindow* window, int w, int h)
+{
+    win_width = w; win_height = h;
```
xlz: Here, coding style please. Do not hide two statements in one line.

examples/viewer.cpp
```diff
@@ -143,17 +158,24 @@ bool Viewer::render()
     {
         libfreenect2::Frame* frame = iter->second;
 
-        glViewport(x, y, width, height);
-        x += width;
-        if (x >= 1024)
+        glViewport(x, y, win_width, win_height);
```
xlz: The bug is here. `win_width` is used for the width of the total window. There are 4 viewports in the windows, each of which should be half of the size. `width_width/2` here and so on for the rest of the code.

xlz:
> Tested only on Mac with retail Kinect v2 in 'cl' mode.

@RyanGordon's test should also be on a Mac. So your test should not have worked?! How did it work?

examples/viewer.cpp
RyanGordon: Since the width here for each viewport was 600, shouldn't the total window width be 1200, and not 1280?

floe:
I'll close this in favor of #486 (which contains original commits by @mbgeek anyway). Please test/discuss the new PR.

PR #469, opened 2015-11-16
quick README fix

rahulraw:
I'm guessing you want users to have build in the libfreenect/ directory instead of the libfreenect/depends/ directory.


floe:
Right, if you really execute everything verbatim, this is needed. Thanks.

PR #470, opened 2015-11-17
Update depth_packet_processor.cpp

vinouz:
Changed gaussian kernel coefficients so that total is 1.0f (was 0.9999999f)


xlz:
The fix is correct. But please move "// sum shall be 1.0f, not 0.9999999f ? (rounding errors accumulating here)" out of the comment and put it in the commit message (also write better commit title).

floe:
Already merged, so a little commit pollution here, but not too bad IMHO. Thanks!

Issue #109, opened 2014-12-02
Using OpenCL with Intel GPU on Linux


wiedemeyer:
Thanks to the work of @masterhui, we have instructions to get OpenCL running on Intel GPUs using Beignet. I think this is useful for many people here too. The original thread is https://github.com/code-iai/iai_kinect2/issues/11.

#### Known configuration
- Ubuntu 14.04
- Kernel 3.13 (>= 3.13.0-35-generic) or Kernel 3.16 (needed for the Intel USB 3.0 Controller)
- Beignet v1.0 (http://www.freedesktop.org/wiki/Software/Beignet/)

#### Dependencies for Beignet
For Beignet the following depencies have to be installed manually:
* ocl-icd-dev
* ocl-icd-libopencl1
* libdrm / libdrm-dev
* llvm-3.5 / llvm-3.5-dev
* clang-3.5 / clang-3.5-dev
* libegl1-mesa-dev
* libedit-dev

#### Building Beignet
Download and compile the Beignet v1.0 release from source (there is a Beignet_v0.3 binary for Trusty, but it is very old, buggy and slow)

##### Additional steps (if needed):

* Error "clang: not found":

  ```
sudo ln -s /usr/lib/llvm-3.5/bin/clang /usr/bin/clang
```
* Known Beignet issue with Kernel 3.15/3.16 (see Beignet readme); fix is to disable cmd_parser:

  ```
sudo su
echo 0 > /sys/module/i915/parameters/enable_cmd_parser
```

* To get 100% pass rate on the Beignet unit tests you may have to:
  * Execute directly on hw: ssh-session might not work
  * Execute as root

*Note: Both previous points have to to with the fact that no x-server was installed. Apparently this will be fixed in a future release of Beignet.*

#### Results on Intel i7-3840QM (mobile hardware)
* **~100 fps** on the OpenCLDepthPacketProcessor (compared to **< 5 fps** on same hardware using CPU-based depth registration!)

  ```
...
[OpenCLDepthPacketProcessor] avg. time: 10.1716ms -> ~98.3129Hz
[TurboJpegRgbPacketProcessor] avg. time: 16.0787ms -> ~62.194Hz
...
```

spacetrain:
Hello,
I followed your instruction, and the compile is successful.
But Protonect is now segfault. I thought it was the problem of the global variable 'shutdown', but I figured out it has already been fixed by replacing the name to 'will_shutdown'.
So I dbged, and the output is as follows:

[...]

PR #472, opened 2015-11-18
Add some OpenCL instructions to README

xlz:
This should include the last remaining issues in #109.

One is left unsolved: what to do with beignet-dev 1.1.1 on Ubuntu 14.04 which is required for Broadwell Iris graphics. Choices: let user build it; put it in floe's ppa; I create a ppa.

> As noted above, beignet 1.0.3 does not support Iris 6100 BroadWell. That commit is included in beignet 1.1.1. Probably want to find a way to provide 1.1.1 binary and do not force users to build from source. Ubuntu xenial has beignet-dev 1.1.1, but this binary has dependency on libllvm3.5v5 which is after gcc5 transition and can't bt used in trusty. A ppa is required to backport beignet-dev to trusty.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ sudo apt-get install build-essential cmake pkg-config libturbojpeg libjpeg-turbo
 
 1. Install libusb. `sudo apt-get install libusb-1.0-0-dev`. The version must be >=1.0.20. You may use `sudo apt-add-repository ppa:floe/libusb` if version 1.0.20 is not available.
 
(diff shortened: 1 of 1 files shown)


floe:
I can put it in my PPA if it's possible to build it with the regular llvm* packages from 14.04?

xlz:
@floe libllvm3.5 is in 14.04 LTS and is fine. The problem with later release is they are after gcc5 abi transition (and become libllvm3.5v5 and libllvm3.6v5).

While you are at it, can you also build ocl-icd 2.2.4+ for the deadlock bug? It is small with few dependencies. Interestingly you can also edit `control` file in beignet-dev and have it depend on `ocl-icd-opencl-dev (>= 2.2.4)`, then buggy version of ocl-icd will get automatically updated.

PR #476, opened 2015-11-20
API cleanup/refactoring

xlz:
I was reviewing and writing up API documentation. It is time to minimize and shape API well before a major release. After 0.1, minor version will have to guarantee API compatibility and ABI compatibility for binary package release. Removing API after major release is very hard.

This PR removes packet processors from public APIs, and hides a few private functions from exported symbols. Exported symbols are greatly reduced. 

I try my best to ensure no API breakage (ABI has to change). Reviews and suggestions are welcome. Each commit is properly functionally organized for review. It is possible more can be done to ensure future proof ABI compatibility.

Each commit is fully tested on Debian and Ubuntu. iai_kinect2 with ROS Indigo is also tested.

Interestingly this commit `api: Work around setConfiguration in iai_kinect2` does not change ABI at all. Reverting it later would be safe.


floe:
Generally a good idea. I'm unsure if it would actually make sense to leave the internals of `Registration` exposed so people don't have to hack around this if they want to access the internal tables - I think that might be useful for some CV scenarios?

xlz:
Exposing those tables directly makes it hard to maintain ABI compatibility. Though adding a function to return a pointer to one of them would be ok.

But I'm not sure of the usefulness of these tables. They are optimization cache of `distort()` and `depth_to_color()` and are hard to understand. If something here needs to be exposed, it can be `distort()` and `depth_to_color()`. I can add them. What do you think?

floe:
OK, I guess you're right, this is not something the average user needs to be bothered with. It's too specific to libfreenect2. So this can be merged as-is.

xlz:
Rebased.

PR #477, opened 2015-11-20
API documentation

xlz:
Again, only the last two commits are for this PR.

Review generated sample documentation here http://xlz.github.io/libfreenect2/. Feedback and critique are welcome. I also added some styles to doxygen to try to make it look less ugly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -257,10 +257,10 @@ INSTALL(FILES "${PROJECT_BINARY_DIR}/freenect2.pc" DESTINATION lib/pkgconfig/)
 
 find_package(Doxygen)
 IF(DOXYGEN_FOUND)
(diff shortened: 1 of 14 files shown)


floe:
+1 great work, thanks! Docs look good to me. I don't currently want to go the whole auto-generation route using Jenkins etc., should be sufficient to generate them locally once and put them into gh-pages branch AFAICT?

xlz:
Yes.

xlz:
Rebased. You can genenerate the docs with version number by `cmake .. -DPROJECT_VERSION=0.1; make doc`. (Ignore the warnings CMP0048)

I don't have a definite idea on how to set up the versioning system right now. That is left for later.

xlz:
Rebased to #479 (one more change is needed for stable ABI).

xlz:
Rebased and tested.

PR #479, opened 2015-11-22
Forward ABI compatibility of Frame

xlz:
`Frame` right now always allocates memory and there is no way to turn it off.

In the future, VT, VAAPI, and Tegra JPEG decoder, and CUDA processor all need to use custom memory allocators, and they will have to break ABI without this.

Example of how VT decoder changes it: https://github.com/OpenKinect/libfreenect2/pull/365/files#diff-7bd5995c21c16482ff1a19a27816180dL58


RyanGordon:
Just tested this on my MacBook Air and it compiles and runs well :+1:

PR #486, opened 2015-11-29
Viewer Scaling Fix

RyanGordon:
This makes it so resizing the viewer, will properly scale the 4 viewports inside.

Supersedes https://github.com/OpenKinect/libfreenect2/pull/467


pookiefoof:
Hi, 

To work properly on Macs with Retina displays (where window.size != framebuffer.size), I think when setting the viewport, "win_width" and "win_height" should be replaced by the width/height value returned by glfwGetFramebufferSize(), e.g., win_width -> framebuffer.width / 2.

RyanGordon:
Hi @pookiefoof,

It looks like glfwGetFramebufferSize() requires window as the first argument, so we must call glfwCreateWindow first with win_width / win_height. Do you have an example patch that you could show us?

pookiefoof:
Hi @RyanGordon,

Yes, we can still use win_width / win_height to create the window; but in Viewer::render(), instead of setting the viewport using win_width / win_height, we should change it to:

```c++
        int fb_width, fb_height;
        glfwGetFramebufferSize(window, &fb_width, &fb_height);
        glViewport(x, y, fb_width / 2, fb_height / 2);

        x += (fb_width / 2);
        if (x >= (fb_width))
        {
            x = 0;
            y += (fb_height / 2);
        }
```

RyanGordon:
Hi @pookiefoof,

Thanks for the patch. I ran into an issue with the IR viewport disappearing while resizing on my Mac, and it appears to be a rounding issue with the `fb_width / 2` acting as a floor with integer precision rounding, while the previous logic usedthe  `if (x >= (win_width * 2))` line which doesn't suffer from the division rounding error.

I think probably using `(fb_width + 1) / 2` is probably the more efficient option versus `ceil(fb_width / 2.0)`

RyanGordon:
I've updated the patch for retina screens here: https://github.com/RyanGordon/libfreenect2/commit/ad5666a5689083cbeac686aafe3b16b1b61d90a3. I've tested it on my non-retina MacBook Air and it seems to work well.

@pookiefoof Can you give this a go on your retina and see if it fixes the issue?

Also, can someone test this on Windows and Linux to ensure it works for them as well?

pookiefoof:
Hi @RyanGordon,

Yes it was a stupid mistake :P. I would suggest just change it to: `if (x >= (fb_width - 1))`, since your current solution will probably crop the image a *little* bit... But really it is no big deal.

xlz:
Tests OK on Debian and Ubuntu 14.04.

I can at least point out several more imperfections (flashing when resizing, viewports get switched for on e frame). The current implementation for an example viewer is acceptable.

RyanGordon:
@pookiefoof I'll go ahead and change that line to use fb_width - 1 to solve the slight cropping

@xlz Thanks for testing! Is there any quick fixes you could suggest for those two issues?

xlz:
I don't see anything easy fix atm. There is a lot potential improvement in viewer.cpp, but I don't think it's worth the time, because it's just a viewer.

RyanGordon:
Makes sense to me, I pushed the last small fix in 188f6bb and it looks like it's good to go @floe

floe:
I'll take your word for it :-)

Issue #489, opened 2015-12-01
Error LNK1120: 1 resolved external symbol


jaebums:
Hello, I have been trying to build the library but have failed over 1 week. I do not have a good experience solving this kind issues partly since I have been using C but not C++. Any comment how I can fix this kind of error?

![image 1](https://cloud.githubusercontent.com/assets/16108252/11515794/bd90d4a6-984d-11e5-94f0-4fb64ac00536.png)

xlz:
It's a bug. I didn't test it on Windows.

MSVC refuses to resolve the symbol when the return type is different (a different type of pointer). I didn't see this with GCC. I'll create a fix promptly.

xlz:
Can you try the above PR 490?

jaebums:
0 Warning, 0 Error! Thank you so much!

PR #427, opened 2015-10-15
add (untested) SWIG interface file for Java bindings

floe:
Starting point for fixing #417, not yet finished & untested. /cc @jonnius @xlz

PR #490, opened 2015-12-02
Fix MSVC FTBFS

xlz:
The workaround cdd4f06 (from #476) broke MSVC building (reported in #489). I didn't test MSVC in #476. MSVC refuses to resolve the symbol because the return type is different, which was the point of the workaround.

Alternative workarounds would make it more a mess. I have sent a patch to iai_kinect2 directly to use new API. https://github.com/code-iai/iai_kinect2/pull/189

Build tested OK with MSVC and GCC after this PR.

### DIFF
--- a/include/internal/libfreenect2/depth_packet_processor.h
+++ b/include/internal/libfreenect2/depth_packet_processor.h
@@ -33,7 +33,6 @@
 #include <stdint.h>
 
 #include <libfreenect2/config.h>
-#define LIBFREENECT2_SETCONFIGURATION_COMPAT_INTERNAL
 #include <libfreenect2/libfreenect2.hpp>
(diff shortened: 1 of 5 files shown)


RyanGordon:
This also fixes an issue with generating SWIG code on master in https://github.com/OpenKinect/libfreenect2/pull/427 :+1:

floe:
@RyanGordon great, so did you also test this on MSVC? Then I think I'll merge it right away, multiple inheritance is actually a big problem with SWIG IIRC, so good to get rid of it.

RyanGordon:
@floe Unfortunately, I don't have MSVC setup to test this

xlz:
#489 user reported this worked on Windows.

PR #494, opened 2015-12-03
Mostly usb fixes

xlz:
* An OpenCL usability fix.
* USB warmup protocol and pause

Tested on Debian and Ubuntu 14.04. Will test with Jetson and Windows later.

### DIFF
--- a/examples/Protonect.cpp
+++ b/examples/Protonect.cpp
@@ -48,6 +48,27 @@ void sigint_handler(int s)
   protonect_shutdown = true;
 }
 
+bool protonect_paused = false;
+libfreenect2::Freenect2Device *devtopause;
+
+//Doing non-trivial things in signal handler is bad. If you want to pause,
+//do it in another thread.
+//Though libusb operations are generally thread safe, I cannot guarantee
+//everything above is thread safe when calling start()/stop() while
+//waitForNewFrame().
+void sigusr1_handler(int s)
+{
+  if (devtopause == 0)
+    return;
+/// [pause]
+  if (protonect_paused)
+    devtopause->start();
+  else
+    devtopause->stop();
+  protonect_paused = !protonect_paused;
+/// [pause]
+}
+
 //The following demostrates how to create a custom logger
 /// [logger]
 #include <fstream>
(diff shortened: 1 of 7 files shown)


examples/Protonect.cpp
```diff
@@ -48,6 +48,27 @@ void sigint_handler(int s)
...
+//do it in another thread.
+//Though libusb operations are generally thread safe, I cannot guarantee
+//everything above is thread safe when calling start()/stop() while
+//waitForNewFrame().
+void sigusr1_handler(int s)
+{
+  if (devtopause == 0)
+    return;
+/// [pause]
+  if (protonect_paused)
```
RyanGordon: Should the pausing be moved out of the signal handler and into the main loop in Protonect.cpp?

xlz: After pausing, the main loop will `waitForNewFrame()` forever. So no.

RyanGordon: Ah right, makes sense. Do we care enough in this file to pull it out in a separate thread then?

xlz: I don't have time to implement it at the moment just for an example.

I think a good solution is to use waitForNewFrame(frame, milliseconds), which needs `wait_for()`, a C++11 feature. This would need backporting `wait_for()` to tinythread. I have no idea how hard that it.

xlz: I looked at it more closely. `wait_for` is actually quite non-trivial to get right.

One other alternative is to catch the signal in the main loop and pause right there until one more signal, and then continue the main loop. But this still needs a bunch of `#ifdef` to use platform dependent timing functions or signal functions.

ahundt:
Don't know if there is a [boost](boost.org) dependency or you're avoiding one, but boost does provide this functionality.

examples/Protonect.cpp
ahundt: Don't know if there is a [boost](boost.org) dependency or you're avoiding one, but boost does provide this functionality.

xlz: Boost dependency is being avoided. See https://github.com/OpenKinect/libfreenect2/pull/14

RyanGordon:
Just tested on my Mac again and it's working very well

floe:
Great - once I get a Windows test report, I'll merge it right away.

xlz:
I'll post test results next week. I brought test machines today but forgot to bring HDMI cables to connect them to monitors.

xlz:
Caught and fixed one bug: SIGUSR1 is undefined on Windows and causes FTBFS.

Now tested on Jetson Ubuntu 14.04 and Windows 7/8 MSVC 2013/5.

floe:
Excellent, thanks.

Issue #498, opened 2015-12-08
GL memory leak in viewer.cpp


RyanGordon:
I've observed that there is a memory leak in viewer.cpp that grows unboundedly over time. Over the course of a few minutes, the memory usage grows to 200+MB:

![Memory Call Stack](https://cldup.com/uoZWxc3pFt-3000x3000.png)

Specifically the 3 functions that are leaking memory are:
```
gl()->glBindVertexArray(triangle_vao);
gl()->glBindBuffer(GL_ARRAY_BUFFER, triangle_vbo);
gl()->glBufferData(GL_ARRAY_BUFFER, sizeof(vertices), vertices, GL_STATIC_DRAW);
```

I haven't looked into why this is or how to fix it but I assume there's some type of deallocation we need to do for those functions

PR #499, opened 2015-12-09
Fix Memory Leak in Viewer.cpp

RyanGordon:
Deallocates VAO and VBO in viewer.cpp so that memory doesn't leak within the GL library.

Fixes https://github.com/OpenKinect/libfreenect2/issues/498

### DIFF
--- a/examples/viewer.cpp
+++ b/examples/viewer.cpp
@@ -225,6 +225,9 @@ bool Viewer::render()
             glDrawArrays(GL_TRIANGLES, 0, 6);
             ir.deallocate();
         }
+
+        gl()->glDeleteBuffers(1, &triangle_vbo);
+        gl()->glDeleteVertexArrays(1, &triangle_vao);
     }
 
     // put the stuff we've been drawing onto the display


xlz:
Looks reasonable enough, and should fix other platforms too (but only tests can prove).

I may later run it through valgrind and see if there are any general memory leaks.

pookiefoof:
Didn't even notice that... Actually there's no need to create/delete VBO/VAO per frame, just create them during initialization, and *bind* them each time you want to use.

xlz:
There is no need to update the vertex definitions too. And there is no need to use two triangles, just one is enough for one texture. These are optimizations, and are reserved for 0.2 at this point near release.

RyanGordon:
@pookiefoof @xlz Would one of you like to create a follow-up PR with those optimizations for 0.2?

Separately, I tried running it through Valgrind as well but it seemed to slow down the process to the point that the USB communication wasn't working properly and it'd never make it into the main loop and so I couldn't get any useful information from it. Maybe that's just my low-powered mac though. Eventually I ended up running it through the mac utility iprofiler, which has a nice GUI for viewing the data, and I couldn't see any other memory leaks at all under the opengl processor. We however malloc about 49MB for the Texture's in the depth processor which stick around for the entire process but doesn't grow or shrink. I didn't look at it very closely but I don't think we can deallocate any of those texture's.

xlz:
I have some ideas on how to improve it. It's lower priority than the core part though.

floe:
Any other test reports? It looks completely OK, just would like to be sure.

xlz:
Tested on Jetson Ubuntu 14.04 and Windows 7/8 MSVC 2013/5.

RyanGordon:
Awesome thanks :+1:

PR #507, opened 2015-12-16
Preemptive API expansion

xlz:
* Add return values to `Freenect2Device start(), stop(), close()` so device errors can be reported.
* Add a status field to `Frame` so processor errors can be reported. And a new `format` field. Also some minor tweaking in field ordering.

Though no actual error reporting is implemented. This only extends the API to preemptively avoid ABI breakage between 0.1 and 0.2.

This changes the current API though but most applications should be able to run without changes.

Tested with Debian, Ubuntu 14.04, Windows 8/MSVC2015, and tested with iai_kinect2.

### DIFF
--- a/include/libfreenect2/frame_listener.hpp
+++ b/include/libfreenect2/frame_listener.hpp
@@ -44,7 +44,7 @@ namespace libfreenect2
 class LIBFREENECT2_API Frame
 {
   public:
-  /** Available types of frames and their pixel format. */
+  /** Available types of frames. */
   enum Type
   {
     Color = 1, ///< 1920x1080 32-bit BGRX.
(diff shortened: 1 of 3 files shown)


RyanGordon:
Tested on MacBook Air 10.11.2 - Working fine as well

floe:
This looks straightforward enough, thanks. Merging.

ahundt:
I think there are some issues here, a uint32_t 0.1 ms timestamp is could be too small, it would wrap in only 49 days! I know that seems like a lot but if someone just sits it on a wall watching for passerbys it could easily get that high in <2 months once the code is more hardened.

(2^32) milliseconds =
49.7102696 days

The sequence number is probably too small as well:
((2^32) / 30) * seconds =
1 657.00899 days

Perhaps it would be best to use the standard combo on unix systems, which I believe is long sec + long nanosec since the epoch in 1970? Could also be time since program startup.


As for the format, should different orders and types be considered? For example RGBD could all be float values, or 3 chars and a float, or some other appropriate combination. [gil does this in a clever way](http://www.boost.org/doc/libs/1_60_0/libs/gil/doc/html/gildesignguide.html) but perhaps that is too complicated, either way should other storage orders be accounted for?

floe:
I think the `uint32_t` for timestamp and sequence is simply what the Kinect itself sends in the packet footer, so no way to change it.

ahundt:
Oh I see anything based on the actual hardware should definitely stay true to form. I thought it was going to be data added by the library. Kind of makes me want to leave an xbox with kinect running for 2 months to see if any shenanigans occur.

Issue #519, opened 2016-01-01
minor fix for mac OSX 10.8.5


robozo:
Protonect failed with GLFW error, "The requested client API version is unavailable".
Turns out that OSX 10.8.5 only supports OpenGL version 3.2, and GLFW asked for 3.3.

So in viewer.cpp and opengl_depth_packet_processor.cpp, changed
glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 3);  to
glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 2);
things now working.

Although sometimes tricky to get Kinect recognized, have to reboot or plug/unplug
as described in FAQ.  Protonect runs at 40 Hz or so, mid-2012 macbook pro.

xlz:
Can you make a pull request?

robozo:
Hi, am new to github, didn't know about pull requests etc.
Mostly just wanted to leave a note to save someone some time, in case they run across
the same problem.  Will "close" this now, and maybe do a pull request later
when I have some time to figure out how to do it...

PR #520, opened 2016-01-02
opengl: Lower version to 3.2 for older Mac OSX

xlz:
Proposed by @robozo in #519.


floe:
No side effects on current MacOS?

xlz:
According to GLFW, `glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 2)`

> give you a context for the highest version of OpenGL greater than or equal to 3.2 that is supported by your OS and GPU.

So in principle no side effect. @RyanGordon can test this.

xlz:
Tested: Debian, Ubuntu 14.04 with iai_kinect2, Windowss 8.1, Ubuntu 14.04 on Jetson.

RyanGordon:
Sorry for the delay, I can confirm this is indeed working on OSX 10.11.2

PR #530, opened 2016-01-07
Release cleanup

xlz:
* Re-organize docs and misc files to libfreenect model
* Add versioning (libfreenect2.so -> libfreenect2.so.0.1.0)
* Fix static analysis warnings

The libfreenect2.so.0.1.0 change has some potential effect on downstream iai_kinect2 etc. I'll test these together.


xlz:
There are a few more things that I'm not sure if needed after this PR.
* A standalone LICENSE file to describe the license of this project, and attribute third party licenses. Right now it is OK though because the licenses are written in the headers, but binary release need a standalone file.
* Binary release specific documentation. INSTALL.txt: manual installation of VC++ runtime, dependency on OpenCL.dll, libusbK driver. examples/HOWTO.txt.
* Possible usbdk replacement for libusbK, if that improves things for Windows 7 users.
* I guess there are some issues with freenect2.pc.in and freenect2.cmake.in that are specific to distro policies, but without feedback of packagers I don't know exactly.
* What to do with static library. There is nominal build support, but I never tested it and do not know how static build can be used.

xlz:
Tested: Debian, Ubuntu 14.04 with iai_kinect2, Windowss 8.1, Ubuntu 14.04 on Jetson.

xlz:
Add this packaging stuff which has no effect on testing results.

RyanGordon:
Tested all 3 Protonect modes on Mac OSX 10.11 and everything appears to be working

floe:
Great, all platforms done! I think that's really the last one for 0.1. Thanks everyone, and @xlz in particular. Merging.

Issue #548, opened 2016-02-01
Impossible to process depth image without processing color image


matthieu-ft:
in the Registration module, I would like to be able to work only with depth and not to have to process the color image. Indeed, if I want to get any 3D Point associated with a given pixel value, the current implementation forces me to register the color image, which is time consuming and critical for my current application (I make projections on bodies that I detect using the depth image of the kinect).

I implemented two additional methods for that. See the pull request bellow.

xlz:
To follow in #549.

PR #549, opened 2016-02-01
registration: Add depth-only methods

matthieu-ft:
- undistortDepth() is the equivalent for apply() but without color
- getPointXYZ() is the equivalent for getPointXYZRGB() without color

this is the solution that I suggest for #548

### DIFF
--- a/src/registration.cpp
+++ b/src/registration.cpp
@@ -50,7 +50,9 @@ public:
 
   void apply(int dx, int dy, float dz, float& cx, float &cy) const;
   void apply(const Frame* rgb, const Frame* depth, Frame* undistorted, Frame* registered, const bool enable_filter, Frame* bigdepth, int* color_depth_map) const;
+  void undistortDepth(const Frame *depth, Frame *undistorted) const;
   void getPointXYZRGB (const Frame* undistorted, const Frame* registered, int r, int c, float& x, float& y, float& z, float& rgb) const;
+  void getPointXYZ (const Frame* undistorted, int r, int c, float& x, float& y, float& z) const;
   void distort(int mx, int my, float& dx, float& dy) const;
   void depth_to_color(float mx, float my, float& rx, float& ry) const;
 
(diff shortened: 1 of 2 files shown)


xlz:
Generally OK. Will review closely later.

Can you move the rationale from #548 to the commit message? It's good to have descriptive commit messages. Also, please use the subject format `registration: Add depth-only methods`.

src/registration.cpp
```diff
@@ -123,6 +125,8 @@ void Registration::apply(const Frame *rgb, const Frame *depth, Frame *undistorte
   impl_->apply(rgb, depth, undistorted, registered, enable_filter, bigdepth, color_depth_map);
 }
 
+
+
```
xlz: Two new lines?

src/registration.cpp
```diff
@@ -301,6 +350,34 @@ void RegistrationImpl::getPointXYZRGB (const Frame* undistorted, const Frame* re
   }
 }
 
+void Registration::getPointXYZ(const Frame *undistorted, int r, int c, float &x, float &y, float &z) const
+{
+    impl_->getPointXYZ(undistorted,r,c,x,y,z);
+}
+
```
xlz: I think it can be implemented like this:
```cpp
void Registration::getPointXYZ (const Frame* undistorted, int r, int c, float& x, float& y, float& z) const
{
  float dummy;
  impl_->getPointXYZRGB(undistorted, undistorted, r, c, x, y, z, dummy);
}
```
What do you think?

xlz:
I mean using "registration: Add depth-only methods" in the commit subject.

src/registration.cpp
matthieu-ft: I do agree that we have to get rid of code duplication, but I prefer to do the other way around, which is to say implement getPointXYZRGB using getPointXYZ. I have updated my commit, have a look and let me know what you think!

xlz:
My review is complete.

This needs a test as follows:

1. Take a pair of rgb and depth frames
2. Use `apply()` to get `undistorted1`
3. Use `undistortDepth()` to get `undistorted2`
4. Verify `undistorted1` and `undistorted2` are identical.

If you can provide the test result it'd be great. After that I'll give it a simple test.

matthieu-ft:
Thank you! So basically, I do this on my side and you tell me if this worked?
I will do this tomorrow.

xlz:
You would post the test results to show the code works correctly.

matthieu-ft:
So I made the following test for 3000 different frames, and I can confirm that both undistorted depth frames are identical!

    float cumulNorm = 0;
    float cumulDiff = 0;

    registration->apply(rgb, depth, &undistorted_apply, &registered);
    registration->undistortDepth(depth,&undistorted_alone);

    //we are going to compare the frames
    float* alone_ptr = (float*)undistorted_alone.data;
    float* apply_ptr = (float*)undistorted_apply.data;
    for(size_t i =0; i<512*424; i++)
    {
        //with cumulNorm I checked that the frames are not empty
        cumulNorm += abs(*apply_ptr);
        cumulDiff += abs(*apply_ptr - *alone_ptr);

        apply_ptr++;
        alone_ptr++;
    }
    
    if(cumulDiff > 0 )
    {
      nbBadRatios ++;
    }

Issue #438, opened 2015-10-21
kinect data logging aka recording and playback?


ahundt:
Is it possible to record kinect data to a file and play it back?

I'm calling this data logging but don't want to confuse it with the status/error logging components of the library I found in my searching for support.

xlz:
ROS can do this. I can't think of a use case of playback where there is no ROS.

Record and export to PCL would be a useful addition to examples.

svddries:
I also think this would be very valuable, even when not using ROS or pcl (I'm currently working on a kinect project in which neither of them are used).

I guess the easiest way would be to store the raw rgb data packages obtained from the device (which are nicely JPG compressed) and either store the depth values for the depth stream directly, or compress them using PNG (which seems to give good compression results, in my experience).

davudadiguezel:
It is possibe. You can use NiViewer that comes with openni2 to record. You'll get a .oni file that you can playback. But if you want to do skeleton detection with Nite2, you'll have to overcome some bugs in Nite2 and I don't know, if somebody has ever did this on Linux. If there is somebody please let me know. But recording and playback is with openni2 no problem.

ahundt:
I don't need skeleton detection, just have to log the data stream because not everyone in my group has usb3 support, plus it would be able to test new code away from the physical rig just using a log of the data. I'm not using ROS and hope to remain cross platform and minimize overhead so I don't think I should necessarily start using it either, unfortunately.

xlz:
Raw JPEG images are 20MB/s. "Raw" 32-bit float depth images are 25MB/s. Do you think you need compression?

ahundt:
Hmm, at that data rate it could be tough to log to a non-ssd without compression so it would probably be a good option.

TheodoreT:
@ahundt did you manage to store raw data?

ahundt:
Right now I'm just saving single frame point clouds out with this: https://github.com/ahundt/libfreenect2pclgrabber

A real implementation would likely still prove useful.

TheodoreT:
@ahundt thanks for the feedback, I see. Actually, at the moment I am saving individual frames as well. I am using opencv to convert libfreenect2::Frames into opencv cv::Mat and then either to save as CV_16UC1 .png files or as CV_32FC1 .exr files. However, though I prefer the 16bit .png files due to the file size I am facing the problem I am describing at the #509 with the bigdepth image and the white background if I try to restore it.

ahundt:
[...]

PR #551, opened 2016-02-01
Add a dump depth processor.  Reactivate the RGB dump processor.

brendandburns:
Add a dump pipeline.

### DIFF
--- a/include/internal/libfreenect2/depth_packet_processor.h
+++ b/include/internal/libfreenect2/depth_packet_processor.h
@@ -175,6 +175,36 @@ public:
 private:
   OpenCLDepthPacketProcessorImpl *impl_;
 };
+
 #endif // LIBFREENECT2_WITH_OPENCL_SUPPORT
+
+class DumpDepthPacketProcessor : public DepthPacketProcessor
+{
+ public:
+  DumpDepthPacketProcessor();
+  virtual ~DumpDepthPacketProcessor();
+
+  virtual void process(const DepthPacket &packet);
+
+  virtual void loadP0TablesFromCommandResponse(unsigned char* buffer, size_t buffer_length);
+  virtual void loadXZTables(const float *xtable, const float *ztable);
+  virtual void loadLookupTable(const short *lut);
+
+  const unsigned char* getP0Tables();
+
+  const float* getXTable();
+  const float* getZTable();
+
+  const short* getLookupTable();
+
+ protected:
+  unsigned char* p0table_;
+
+  float* xtable_;
+  float* ztable_;
+  
+  short* lut_;
+};
+
 } /* namespace libfreenect2 */
 #endif /* DEPTH_PACKET_PROCESSOR_H_ */
--- a/include/libfreenect2/frame_listener.hpp
+++ b/include/libfreenect2/frame_listener.hpp
@@ -58,12 +58,14 @@ class LIBFREENECT2_API Frame
     BGRX,
     RGBX,
     Gray,
-    Float
+    Float,
+    // Note that if format is 'raw' then 'bytes_per_pixel' actually contains the number of bytes
+    Raw
   };
 
   size_t width;           ///< Length of a line (in pixels).
   size_t height;          ///< Number of lines in the frame.
-  size_t bytes_per_pixel; ///< Number of bytes in a pixel.
+  size_t bytes_per_pixel; ///< Number of bytes in a pixel. If frame format is 'Raw' this is the buffer size.
   unsigned char* data;    ///< Data of the frame (aligned). @see See Frame::Type for pixel format.
   uint32_t timestamp;     ///< Unit: roughly or exactly 0.1 millisecond
   uint32_t sequence;      ///< Increasing frame sequence number
--- a/include/libfreenect2/packet_pipeline.h
+++ b/include/libfreenect2/packet_pipeline.h
@@ -66,6 +66,13 @@ protected:
   PacketPipelineComponents *comp_;
 };
 
+ class LIBFREENECT2_API DumpPacketPipeline: public PacketPipeline
+ {
+ public:
+   DumpPacketPipeline();
+   virtual ~DumpPacketPipeline();
+ };
+
 /** Pipeline with CPU depth processing. */
 class LIBFREENECT2_API CpuPacketPipeline : public PacketPipeline
 {
--- a/src/depth_packet_processor.cpp
+++ b/src/depth_packet_processor.cpp
@@ -29,6 +29,8 @@
 #include <libfreenect2/depth_packet_processor.h>
 #include <libfreenect2/async_packet_processor.h>
 
+#include <cstring>
+
 namespace libfreenect2
 {
 
@@ -98,4 +100,63 @@ void DepthPacketProcessor::setFrameListener(libfreenect2::FrameListener *listene
   listener_ = listener;
 }
 
+DumpDepthPacketProcessor::DumpDepthPacketProcessor()
+  : p0table_(NULL), xtable_(NULL), ztable_(NULL), lut_(NULL) {
+}
+
+DumpDepthPacketProcessor::~DumpDepthPacketProcessor(){
+  delete[] p0table_;
+  delete[] xtable_;
+  delete[] ztable_;
+  delete[] lut_;
+}
+
+void DumpDepthPacketProcessor::process(const DepthPacket &packet) {
+  Frame* depth_frame = new Frame(1, 1, packet.buffer_length);
+  
+  depth_frame->timestamp = packet.timestamp;
+  depth_frame->sequence = packet.sequence;
+  depth_frame->format = Frame::Raw;
+  std::memcpy(depth_frame->data, packet.buffer, packet.buffer_length);
+
+  Frame* ir_frame = new Frame(1, 1, packet.buffer_length, depth_frame->data);
+  ir_frame->timestamp = packet.timestamp;
+  ir_frame->sequence = packet.sequence;
+  ir_frame->data = packet.buffer;
+  ir_frame->format = Frame::Raw;
+
+  if (!listener_->onNewFrame(Frame::Ir, ir_frame)) {
+    delete ir_frame;
+  }
+  ir_frame = NULL;
+  if (!listener_->onNewFrame(Frame::Depth, depth_frame)) {
+    delete depth_frame;
+  }
+  depth_frame = NULL;
+}
+  
+const unsigned char* DumpDepthPacketProcessor::getP0Tables() { return p0table_; }
+  
+const float* DumpDepthPacketProcessor::getXTable() { return xtable_; }
+const float* DumpDepthPacketProcessor::getZTable() { return ztable_; }
+
+const short* DumpDepthPacketProcessor::getLookupTable() { return lut_; }
+
+void DumpDepthPacketProcessor::loadP0TablesFromCommandResponse(unsigned char* buffer, size_t buffer_length) {
+  p0table_ = new unsigned char[buffer_length];
+  std::memcpy(p0table_, buffer, buffer_length);
+}
+  
+void DumpDepthPacketProcessor::loadXZTables(const float *xtable, const float *ztable) {
+  xtable_ = new float[TABLE_SIZE];
+  std::memcpy(xtable_, xtable, TABLE_SIZE * sizeof(float));
+
+  ztable_ = new float[TABLE_SIZE];
+  std::memcpy(ztable_, ztable, TABLE_SIZE * sizeof(float));
+}
+  
+void DumpDepthPacketProcessor::loadLookupTable(const short *lut) {
+  lut_ = new short[LUT_SIZE];
+  std::memcpy(lut_, lut, LUT_SIZE * sizeof(short));
+}
 } /* namespace libfreenect2 */
--- a/src/rgb_packet_processor.cpp
+++ b/src/rgb_packet_processor.cpp
@@ -29,6 +29,7 @@
 #include <libfreenect2/rgb_packet_processor.h>
 #include <libfreenect2/async_packet_processor.h>
 
+#include <cstring>
 #include <fstream>
 #include <string>
 
@@ -49,22 +50,26 @@ void RgbPacketProcessor::setFrameListener(libfreenect2::FrameListener *listener)
   listener_ = listener;
 }
 
-DumpRgbPacketProcessor::DumpRgbPacketProcessor()
-{
-}
-
-DumpRgbPacketProcessor::~DumpRgbPacketProcessor()
-{
-}
+DumpRgbPacketProcessor::DumpRgbPacketProcessor() {}
+DumpRgbPacketProcessor::~DumpRgbPacketProcessor() {}
 
 void DumpRgbPacketProcessor::process(const RgbPacket &packet)
 {
-  //std::stringstream name;
-  //name << packet->sequence << "_" << packet->unknown0 << "_" << jpeg_buffer_length << ".jpeg";
-  //
-  //std::ofstream file(name.str().c_str());
-  //file.write(reinterpret_cast<char *>(packet->jpeg_buffer), jpeg_buffer_length);
-  //file.close();
+  Frame *frame = new Frame(1, 1, 1920*1080*4);
+  frame->sequence = packet.sequence;
+  frame->timestamp = packet.timestamp;
+  frame->exposure = packet.exposure;
+  frame->gain = packet.gain;
+  frame->gamma = packet.gamma;
+  frame->format = Frame::Raw;
+  frame->bytes_per_pixel = packet.jpeg_buffer_length;
+
+  std::memcpy(frame->data, packet.jpeg_buffer, packet.jpeg_buffer_length);
+
+  if (!listener_->onNewFrame(Frame::Color, frame)) {
+    delete frame;
+  }
+  frame = NULL;
 }
 
 } /* namespace libfreenect2 */
--- a/src/packet_pipeline.cpp
+++ b/src/packet_pipeline.cpp
@@ -129,4 +129,13 @@ OpenCLPacketPipeline::OpenCLPacketPipeline(const int deviceId) : deviceId(device
 OpenCLPacketPipeline::~OpenCLPacketPipeline() { }
 #endif // LIBFREENECT2_WITH_OPENCL_SUPPORT
 
+DumpPacketPipeline::DumpPacketPipeline()
+{
+  RgbPacketProcessor *rgb = new DumpRgbPacketProcessor();
+  DepthPacketProcessor *depth = new DumpDepthPacketProcessor();
+  comp_->initialize(rgb, depth);
+}
+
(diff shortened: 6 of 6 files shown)


include/internal/libfreenect2/depth_packet_processor.h
```diff
@@ -175,6 +175,25 @@ class OpenCLDepthPacketProcessor : public DepthPacketProcessor
 private:
   OpenCLDepthPacketProcessorImpl *impl_;
 };
+
 #endif // LIBFREENECT2_WITH_OPENCL_SUPPORT
+
+class DumpDepthPacketProcessor : public DepthPacketProcessor
```
xlz: Can you explain why you want to dump depth packets? The raw depth packets will require P0, X, and Z tables to decode. The tables are pulled from the Kinect at run time. So far my impression is that the raw depth packets are too big to be practical (85MB/s).

src/depth_packet_processor.cpp
```diff
@@ -98,4 +104,51 @@ void DepthPacketProcessor::setFrameListener(libfreenect2::FrameListener *listene
...
+    file.write(reinterpret_cast<char *>(packet.buffer), packet.buffer_length);
+    file.close();
+
+    std::stringstream json;
+    json << "{" << std::endl;
+    json << "\"timestamp\": " << packet.timestamp << "," << std::endl;
+    json << "\"sequence\": " << packet.sequence << "," << std::endl;
+
+    timespec now;
+    clock_gettime(CLOCK_REALTIME, &now);
```
xlz: This function is not portable. This is a pain point in the project so far. There is no portable timing function in C++98 unless you introduce a portability kludge.

We try to use glfwGetTime() or C++11 std::chrono depending on the build config, https://github.com/OpenKinect/libfreenect2/blob/master/src/logging.cpp#L182. Unfortunately this doesn't always work when the user disables OpenGL and does not enable C++11 (some Mac users, some Odroid users).

So two choices for replacing this:

1. Introduce a portable timing function
2. Depend on the build config to use one of glfwGetTime() or C++11 std::chrono, and deal with it when both are unavailable.

include/internal/libfreenect2/depth_packet_processor.h
brendandburns: We'd like to do offline processing of the depth files in order to facilitate faster/easier algorithm refining and testing.

When you mention the tables, can I pull them once and keep them around or are they re-calculated whenever the kinect restarts?

And yes, I'm seeing ~3Mb / frame, so yes, I think your size estimate is correct.  However I'm not offput by those sizes, and I can successfully capture to disk at those rates w/o frame dropping.

src/depth_packet_processor.cpp
brendandburns: thanks for the thorough explanation.  I'll use std::chrono and deal with it being missing.

brendandburns:
Someone on my team suggested that an alternative to this approach would be to extend `libfreenect2::Frame` to include a `Type` of `Raw` and then pass the raw data along via `Frame` objects of type `Raw` to the corresponding `Listener` and have them handle it.  This would probably mean adding a `buffer_length` field to the `Frame` object as well, since the `bytes_per_pixel` isn't applicable for the raw data.

Let me know if you prefer that approach to the one I've implemented here and I can change the implementation.

(and thanks for the fast review!  I will try to get the portable timing function in tomorrow)

include/internal/libfreenect2/depth_packet_processor.h
xlz: OK, I see now storage is not a problem for Google. :)

P0 table contains the initial phase shift of the device and can only be obtained at run time. X and Z tables are algorithmic helper and are derived from the intrinsic parameters of the IR camera. The intrinsic parameters are factory built-in and obtained at run time but you can provide your own calibration parameters.

More details about the tables: https://github.com/OpenKinect/libfreenect2/blob/master/src/libfreenect2.cpp#L63

If you don't plan to do your own calibration, you can save the tables in `DumpDepthPacketProcessor::load*Table***()`.

xlz:
That was what I planned. I planned to add a JPEG format which is produced by a dummy JPEG processor doing nothing. Maybe you don't want to save it to disk, instead you want to stream it over network. We have some demand on a data recording tool #438, and specifying a recording format is not really a job of the library proper.

I'm OK with adding a dump pipeline for debug purposes though, because the code is there and code matters.

I don't want to add a new field if that breaks the ABI. Maybe you can abuse `bytes_per_pixel` for a Raw format by using proper documentation to redefine its meaning. There is quite some cruft in the library unfortunately, but no ABI breakage is accepted.

brendandburns:
ok, SGTM.

I don't think adding a `Type` `Raw` would break the ABI, since `Type` is listed as TBD in the docs.

So how about the following:

We will add `Type` `RAW` to the type enum, and if `Type` is `RAW` `bytes_per_pixel` will actually be `buffer_length` which is kind of hacky but will work.

If that works for you, I will re-arrange this PR to do that.

Thanks
--brendan

xlz:
I meant adding `buffer_length` to `Frame` would break the ABI.

The current version of your plan is good for me.

floe:
Just curious: what's Google doing with the Kinect v2? Anything related to Project Tango? (Feel free to silently ignore me if I'm being too nosy ;-)

brendandburns:
@xlz I've updated this PR to reflect our discussions.  please take another look.

@floe sorry...

--brendan

include/internal/libfreenect2/depth_packet_processor.h
```diff
@@ -175,6 +175,42 @@ class OpenCLDepthPacketProcessor : public DepthPacketProcessor
...
+  virtual ~DumpDepthPacketProcessor();
+
+  virtual void process(const DepthPacket &packet);
+
+  virtual void loadP0TablesFromCommandResponse(unsigned char* buffer, size_t buffer_length);
+  virtual void loadXZTables(const float *xtable, const float *ztable);
+  virtual void loadLookupTable(const short *lut);
+
+  // Get the P0 Tables
+  unsigned char* getP0Tables();
```
xlz: `const unsigned char*`? I think you would not need to modify the values.

include/internal/libfreenect2/depth_packet_processor.h
```diff
@@ -175,6 +175,42 @@ class OpenCLDepthPacketProcessor : public DepthPacketProcessor
...
+  size_t getP0TablesLength();
```
xlz: The length of p0 tables are known, here https://github.com/OpenKinect/libfreenect2/blob/master/include/internal/libfreenect2/protocol/response.h#L220

* `sizeof(P0TablesResponse)`

`loadP0TablesFromCommandResponse()` may get something more than `sizeof(P0TablesResponse)`, but never less.

src/depth_packet_processor.cpp
```diff
@@ -29,6 +29,13 @@
 #include <libfreenect2/depth_packet_processor.h>
 #include <libfreenect2/async_packet_processor.h>
 
+#include <cstring>
+#include <fstream>
+#include <string>
+#include <sstream>
+
+#include <time.h>
```
xlz: Looks like `time.h` is no longer needed.

include/internal/libfreenect2/depth_packet_processor.h
brendandburns: done.

include/internal/libfreenect2/depth_packet_processor.h
brendandburns: removed.

src/depth_packet_processor.cpp
brendandburns: removed (and others too)

brendandburns:
comments addressed, please take another look.

Thanks!
--brendan

include/internal/libfreenect2/depth_packet_processor.h
```diff
@@ -175,6 +175,40 @@ class OpenCLDepthPacketProcessor : public DepthPacketProcessor
...
+
+  virtual void process(const DepthPacket &packet);
+
+  virtual void loadP0TablesFromCommandResponse(unsigned char* buffer, size_t buffer_length);
+  virtual void loadXZTables(const float *xtable, const float *ztable);
+  virtual void loadLookupTable(const short *lut);
+
+  const unsigned char* getP0Tables();
+
+  float* getXTable();
```
xlz: All these `float*` -> `const float*` too.

`short*` -> `const short *`

include/internal/libfreenect2/depth_packet_processor.h
```diff
@@ -175,6 +175,40 @@ class OpenCLDepthPacketProcessor : public DepthPacketProcessor
...
+  float* getZTable();
+
+  short* getLookupTable();
+
+ protected:
+  Frame* depth_frame;
+  Frame* ir_frame;
+
+  unsigned char* p0table_;
+  size_t p0table_length_;
```
xlz: Remove this.

src/depth_packet_processor.cpp
```diff
@@ -98,4 +100,64 @@ void DepthPacketProcessor::setFrameListener(libfreenect2::FrameListener *listene
...
+  delete[] lut_;
+}
+
+void DumpDepthPacketProcessor::process(const DepthPacket &packet) {
+    ir_frame->height = 1;
+    ir_frame->width = 1;
+    ir_frame->bytes_per_pixel = packet.buffer_length;
+    ir_frame->timestamp = packet.timestamp;
+    ir_frame->sequence = packet.sequence;
+    ir_frame->data = packet.buffer;
```
xlz: `packet.buffer` is immediately recycled by the depth stream parser after `process()` ends. `->data` will contain undefined values. You need memcpy here.

My suggestion:
```cpp
...DumpDepthPacketProcessor() : depth_frame(NULL), ...
...
...process(const DepthPacket &packet) {
  depth_frame = new Frame(1, 1, packet.buffer_length); //packet.buffer_length is a fixed value
  depth_frame->timestamp = packet.timestamp;
  depth_frame->sequence = packet.sequence;
  memcpy(depth_frame->data, packet.buffer);
//and make the ir frame a dummy one
  ir_frame = new Frame(1, 1, packet.buffer_length, depth_frame->data);
...
  if (!listener_->onNewFrame(Frame::Depth, depth_frame))
    delete depth_frame; // This way of delete and new can be found in opengl_depth_packet_processor.cpp
  if (!listener_->onNewFrame(Frame::Ir, ir_frame))
    delete ir_frame; // This way of delete and new can be found in opengl_depth_packet_processor.cpp
```

src/depth_packet_processor.cpp
```diff
@@ -98,4 +100,64 @@ void DepthPacketProcessor::setFrameListener(libfreenect2::FrameListener *listene
...
+    }
+    if (listener_->onNewFrame(Frame::Depth, depth_frame)) {
+      depth_frame = new Frame(512, 424, 4);
+    }
+  }
+
+  
+  const unsigned char* DumpDepthPacketProcessor::getP0Tables() { return p0table_; }
+
+  float* DumpDepthPacketProcessor::getXTable() { return xtable_; }
```
xlz: All these changes to const signature, too.

src/rgb_packet_processor.cpp
```diff
@@ -50,21 +52,31 @@ void RgbPacketProcessor::setFrameListener(libfreenect2::FrameListener *listener)
...
-  //
-  //std::ofstream file(name.str().c_str());
-  //file.write(reinterpret_cast<char *>(packet->jpeg_buffer), jpeg_buffer_length);
-  //file.close();
+  frame->sequence = packet.sequence;
+  frame->timestamp = packet.timestamp;
+  frame->width = 1;
+  frame->height = 1;
+  frame->bytes_per_pixel = packet.jpeg_buffer_length;
+  frame->data = packet.jpeg_buffer;
```
xlz: Need memcpy here too. Same reason as in `DumpDepthPacketProcessor::process()`. But I think here it needs allocating with a fixed size `new Frame(1, 1, 1920*1080*4);` (packet.jpeg_buffer_length varies, and using it creates memory fragmentation).

src/rgb_packet_processor.cpp
```diff
@@ -50,21 +52,31 @@ void RgbPacketProcessor::setFrameListener(libfreenect2::FrameListener *listener)
 }
 
 DumpRgbPacketProcessor::DumpRgbPacketProcessor()
+  : frame(new Frame(1920, 1080, 4))
```
xlz: frame(NULL), as suggested similarly in `DumpDepthPacketProcessor`.

src/rgb_packet_processor.cpp
```diff
@@ -50,21 +52,31 @@ void RgbPacketProcessor::setFrameListener(libfreenect2::FrameListener *listener)
...
+  frame->width = 1;
+  frame->height = 1;
+  frame->bytes_per_pixel = packet.jpeg_buffer_length;
+  frame->data = packet.jpeg_buffer;
+
+  frame->exposure = packet.exposure;
+  frame->gain = packet.gain;
+  frame->gamma = packet.gamma;
+
+  if (listener_->onNewFrame(Frame::Color, frame)) {
```
xlz: Change this new and delete order as in `DumpDepthPacketProcessor::process()`.

include/libfreenect2/packet_pipeline.h
```diff
@@ -66,6 +66,13 @@ class LIBFREENECT2_API PacketPipeline
   PacketPipelineComponents *comp_;
 };
 
+ class LIBFREENECT2_API DumpPacketPipeline: public PacketPipeline
+ {
+ public:
+   DumpPacketPipeline();
+   virtual ~DumpPacketPipeline();
+ };
```
xlz: So in the public API you didn't actually expose the interfaces to export the P0, X, Z, LUT tables.

Just add `get*Tables()`? What is your plan?

include/libfreenect2/frame_listener.hpp
```diff
@@ -58,12 +58,14 @@ class LIBFREENECT2_API Frame
     BGRX,
     RGBX,
     Gray,
-    Float
+    Float,
+    // Note that if format is 'raw' then 'bytes_per_pixel' actually contains the number of bytes
+    Raw
```
xlz: It looks like the `Raw` format is not assigned to the frames. Add `xx_frame->format = Frame::Raw` below?

xlz:
My review is complete.

include/libfreenect2/frame_listener.hpp
brendandburns: done.

include/libfreenect2/packet_pipeline.h
brendandburns: Yeah, I will add accessors in a follow up PR.  Does that work?

src/depth_packet_processor.cpp
brendandburns: done.  (thanks for the extensive comment/sample)

I reversed the order of the deletes since the depth_frame owns the data pointer, so we don't want to delete it before handing `ir_frame` off to the depth handler.

src/depth_packet_processor.cpp
brendandburns: done.

src/rgb_packet_processor.cpp
brendandburns: done.

src/rgb_packet_processor.cpp
brendandburns: done.

src/rgb_packet_processor.cpp
brendandburns: done.

brendandburns:
Thanks for the extensive review.  Comments addressed, please re-check.

--brendan

src/rgb_packet_processor.cpp
```diff
@@ -50,21 +53,31 @@ void RgbPacketProcessor::setFrameListener(libfreenect2::FrameListener *listener)
 }
 
 DumpRgbPacketProcessor::DumpRgbPacketProcessor()
+  : frame(NULL)
 {
 }
 
 DumpRgbPacketProcessor::~DumpRgbPacketProcessor()
 {
+  delete frame;
```
xlz: Double free here. As in `DumpDepthPacketProcessor::~DumpDepthPacketProcessor()`, just remove this line.

src/rgb_packet_processor.cpp
```diff
@@ -50,21 +53,31 @@ void RgbPacketProcessor::setFrameListener(libfreenect2::FrameListener *listener)
...
 
 void DumpRgbPacketProcessor::process(const RgbPacket &packet)
 {
-  //std::stringstream name;
-  //name << packet->sequence << "_" << packet->unknown0 << "_" << jpeg_buffer_length << ".jpeg";
-  //
-  //std::ofstream file(name.str().c_str());
-  //file.write(reinterpret_cast<char *>(packet->jpeg_buffer), jpeg_buffer_length);
-  //file.close();
+  frame = new Frame(1, 1, 1920*1080*4);
```
xlz: Still need this:
```
  frame->bytes_per_pixel = packet.jpeg_buffer_length;
```
`1920*1080*4` is the size for the allocation, but not for the actual data.

src/rgb_packet_processor.cpp
brendandburns: yeah, I figured that out ;)  I deleted the member variable entirely, its not needed.

src/rgb_packet_processor.cpp
brendandburns: done.

brendandburns:
thanks, sorry for all the little errors...

--brendan

xlz:
Build test OK.

ABI report looks OK.

```
Functions changes summary: 0 Removed, 0 Changed (5 filtered out), 3 Added functions
Variables changes summary: 0 Removed, 0 Changed, 0 Added variable
Function symbols changes summary: 0 Removed, 0 Added function symbol not referenced by debug info
Variable symbols changes summary: 0 Removed, 3 Added variable symbols not referenced by debug info

3 Added functions:

  'method libfreenect2::DumpPacketPipeline::DumpPacketPipeline()'    {_ZN12libfreenect218DumpPacketPipelineC1Ev, aliases _ZN12libfreenect218DumpPacketPipelineC2Ev}
  'method virtual libfreenect2::DumpPacketPipeline::~DumpPacketPipeline(int)'    {_ZN12libfreenect218DumpPacketPipelineD0Ev}
    note that this adds a new entry to the vtable of class libfreenect2::DumpPacketPipeline
  'method virtual libfreenect2::DumpPacketPipeline::~DumpPacketPipeline(int)'    {_ZN12libfreenect218DumpPacketPipelineD1Ev, aliases _ZN12libfreenect218DumpPacketPipelineD2Ev}
    note that this adds a new entry to the vtable of class libfreenect2::DumpPacketPipeline

3 Added variable symbols not referenced by debug info:

  _ZTIN12libfreenect218DumpPacketPipelineE
  _ZTSN12libfreenect218DumpPacketPipelineE
  _ZTVN12libfreenect218DumpPacketPipelineE

```

brendandburns:
Awesome, thanks!

PR #554, opened 2016-02-03
Add accessors for the various depth tables.

brendandburns:
Follow on to #551 

@xlz I had to add `size_t *len` to the accessors since there is no good way to get the lengths out of the library without moving around things and declaring a new constant for `sizeof(P0TableResponse)`

Let me know if you have a preferred alternate approach.

### DIFF
--- a/include/libfreenect2/packet_pipeline.h
+++ b/include/libfreenect2/packet_pipeline.h
@@ -31,6 +31,8 @@
 
 #include <libfreenect2/config.h>
 
+#include <stdlib.h>
+
 namespace libfreenect2
 {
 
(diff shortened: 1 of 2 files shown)


brendandburns:
that's ok with me, if that's your preference.  It does imply copying the data, but since this is one-off (and not that big) that doesn't bother me.

I can do that if you prefer.

xlz:
Alright then keep the pointers. There is no point in doing the extra copy.

brendandburns:
Comment addressed.  please take another look.

thanks!
--brendan

xlz:
Looks fine. Build tests fine.

PR #555, opened 2016-02-03
Add a simple frame driver.

brendandburns:
@xlz one more (I'm through for the night, I promise ;)

This one might be more controversial, its sort of a sketch.  The intent is to create an equivalent thing that can drive a packet pipeline from files.

Please let me know if you have a preference for how this is structured.

Another approach would be to abstract out a bunch of the packet stuff into some sort of `PacketProvider` interface but that seems like a _way_ bigger refactor.

### DIFF
--- a/include/libfreenect2/libfreenect2.hpp
+++ b/include/libfreenect2/libfreenect2.hpp
@@ -34,6 +34,10 @@
 #include <libfreenect2/packet_pipeline.h>
 #include <string>
 
+namespace tthread {
+  class thread;
+}
+
 namespace libfreenect2
 {
 
@@ -259,6 +263,34 @@ public:
   Freenect2Device *openDefaultDevice(const PacketPipeline *factory);
 private:
   Freenect2Impl *impl_;
+};
+
+class LIBFREENECT2_API Freenect2ReplayDevice {
+ public:
+  Freenect2ReplayDevice(char* directory, PacketPipeline* pipeline);
+ 
+  void setColorFrameListener(FrameListener* listener);
+  void setIrAndDepthFrameListener(FrameListener* listener);
+
+
+  void loadP0Tables(unsigned char* buffer, size_t buffer_length);
+  void loadXZTables(const float *xtable, const float *ztable);
+  void loadLookupTable(const short *lut);
+
+  bool processRawFrame(Frame::Type type, Frame* frame);  
+  void start();
+  void stop();
+  void run();
+
+ protected:
+  void processRgbFrame(Frame* frame);
+  void processDepthFrame(Frame* frame);
+
+ private:
+  PacketPipeline* pipeline_;
+  char* directory_;
+  tthread::thread* t_;
+  bool running_;
 };
 
 ///@}
--- a/src/libfreenect2.cpp
+++ b/src/libfreenect2.cpp
@@ -32,6 +32,10 @@
 #include <libusb.h>
 #include <limits>
 #include <cmath>
+#include <dirent.h>
+#include <iostream>
+#include <fstream>
+
 #define WRITE_LIBUSB_ERROR(__RESULT) libusb_error_name(__RESULT) << " " << libusb_strerror((libusb_error)__RESULT)
 
 #include <libfreenect2/libfreenect2.hpp>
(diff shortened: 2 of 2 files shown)


include/libfreenect2/libfreenect2.hpp
```diff
@@ -261,6 +261,28 @@ class LIBFREENECT2_API Freenect2
   Freenect2Impl *impl_;
 };
 
+class LIBFREENECT2_API Freenect2FrameDriver {
+ public:
+  Freenect2FrameDriver(char* directory, PacketPipeline* pipeline);
```
xlz: `directory` is not used. Is it intended for something else?

xlz:
I don't see any controversies here.

The name Freenect2FrameDriver seems quite generic. Do you intend a replay mechanism, or a fake device? Maybe call it Freenect2ReplayDevice or Freenect2FakeDevice (inspiration from https://github.com/OpenKinect/libfreenect/tree/master/fakenect)?

Looking at the methods you've implemented, it's quite close to an actual Freenect2Device subclass. Have you considered making it an actual Freenect2Device subclass? Then this replay device can be plug-n-play as real devices.

```cpp
dev = freenect2.openDevice("dir:/tmp/capture", new libfreenect2::XXXPipeline);
dev->setColorFrameListener(&listener);
dev->setIrAndDepthFrameListener(&listener);
dev->start();
while(???) {
  listener.waitForNewFrame(frames);
  //.. do things with frames
}
```
So probably you would start a new thread in `dev->start()` to read files and provide them internally to processRawFrame().

Basically I'm hoping for more compatibility/code reuse with `Freenect2Device`. What do you think?

brendandburns:
Ok, I updated this to reflect your comments and bring the API in line with Freenect2Device.

I think we'll want to add a `Freenect2Interface` or somesuch that defines the superset interface that both `Freenect2Device` and `Freenect2ReplayDevice` implement...

This isn't complete, I don't support loading JPEGs yet...

xlz:
Why not `class LIBFREENECT2_API Freenect2ReplayDevice: public Freenect2Device`?

brendandburns:
@xlz `Freenect2Device` has a ton of libusb specific stuff in it, so extending from it seems wrong.  I think a parent interface is the right solution for this kind of thing.

(or pushing the libusb stuff down into a `Freenect2USBDevice`...)

--brendan

xlz:
Can you be more specific about what is libusb specific stuff? I'm not seeing anything particularly limited to libusb. The static consts? `getSerialNumber()`? `getFirmwareVersion()`?

If one method is not applicable to Freenect2ReplayDevice, it can be even implemented like `LOG_DEBUG << "not implemented"; return true;`.

You do want to save and restore the camera parameters with `{get,set}{Color,Ir}CameraParams()` for potential registration.

xlz:
`Freenect2Device` is a pure abstract class. https://github.com/OpenKinect/libfreenect2/blob/master/include/libfreenect2/libfreenect2.hpp#L45

In case you mistook it with `Freenect2DeviceImpl`.

smokhov:
Oh I see another one related to possible Fakenect implementation within libfreenect2. I just reviewed briefly #602 -- where the recording feature is proposed (in addition to streaming). @brendandburns and @xlz what is missing to get this PR in aside from it being diverged now? Possibly jointly with #602 I can have my "Fakenect2" I need. Let me know if I can help.

smokhov:
@xlz 

> Why not class LIBFREENECT2_API Freenect2ReplayDevice: public Freenect2Device?

That made a lot of sense. I've done that part and implementing the methods. I've also re-aligned the API for `stop()`, `run()`, and `start()`. For `getSerialNumber()` and `getFirmwareVersion()` I am returning current `FREENECT2_VERSION` and the API version for the latter.

To test all this do we create an option in `Protonect` to use the replay device instead? I think either that or mod `ProtonectSR` in #602 to do this test. Opinions? Keeping it conservative, I think I will put it in the latter so I could test both recording and replaying in one tool.

smokhov:
> You do want to save and restore the camera parameters with {get,set}{Color,Ir}CameraParams() for potential registration.

For these I am simply replicating `Freenect2DeviceImpl`'s implementation.

smokhov:
From #602 @xlz your concern here was the non-portability of `opendir()` and friends used in this code with Windows. Would something [like this](http://www.two-sdg.demon.co.uk/curbralan/code/dirent/dirent.html) be acceptable to include? Or I will proceed with my refactoring our proposal into the tool instead of core lib.

smokhov:
@brendandburns says:

> This isn't complete, I don't support loading JPEGs yet...

I don't think we want JPEGs now; the implementation currently simply does depth only though some functionality for RGB is there, I think we should work with raw RGB data, for now compatible with the recorder.

xlz:
> To test all this do we create an option in Protonect to use the replay device instead

Protonect was cross-platform tested. To add to Protonect means your feature needs to be cross-platform too. Portability is an issue for that.

The platform independent classes can go into the core library no problem. I would prefer putting platform specific stuff into contrib or tool.

> I don't think we want JPEGs now [...] I think we should work with raw RGB data

I think loading JPEGs is possible through the `Raw` format in `Frame`? JPEG is the raw format and 1920x1080 is the decoded RGB data so I'm not sure what you meant by working with raw RGB data and not wanting JPEGs.

Other than that it looks perfect to me so far.

xlz:
Continue in #942.

PR #556, opened 2016-02-03
Usb error reporting

xlz:
* Properly stop execution when usb errors happen (libusb errors, short command response).
* Use std::vector instead of manual memory management in CommandTransaction

### DIFF
--- a/examples/Protonect.cpp
+++ b/examples/Protonect.cpp
@@ -245,7 +245,8 @@ int main(int argc, char *argv[])
 /// [listeners]
 
 /// [start]
-  dev->start();
+  if (!dev->start())
+    return -1;
 
   std::cout << "device serial: " << dev->getSerialNumber() << std::endl;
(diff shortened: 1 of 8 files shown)


floe:
This does look very sensible. As this would only affect the internal API/ABI, it might be worth considering as bugfix for 0.1.2?

xlz:
Maybe. The bug this is fixing is segfault after usb failure. This merely makes it nicer to handle, but the user still has to stop the entire program upon such errors.

xlz:
v2: `min_response_length` was added to the wrong structure (`CommandData` instead of `CommandBase`) which is sent over USB. This messed up the byte protocol. Fix it.

xlz:
Now tested on Debian and Windows.

@RyanGordon @fran6co I'll merge this PR next week. If you have time, please try it on Mac.

fran6co:
Seems to be working ok on Mac. Didn't get any error to test the full scope of it, though.

xlz:
Manually merged.

PR #557, opened 2016-02-04
Add ability to disable RGB or depth stream

xlz:
Users want to save USB bandwidth and CPU if they don't need RGB or depth.

* Add new `startStreams(bool rgb, bool depth)` to Freenect2Device
* Give an example stream selection with new options `-norgb -nodepth` to Protonect


xlz:
Now tested on Debian and Windows.

@RyanGordon @fran6co I'll merge this PR next week. If you have time, please also try this on Mac (separately).

fran6co:
@xlz works ok, only thing is that if you try to turn off both streams at the same time it just disables the first one. Not a scenario that makes much sense but maybe a warning message or something would be nicer.

xlz:
Manually merged.

PR #563, opened 2016-02-07
Add VA-API JPEG decoder again

xlz:
Previous attempt: #210 (including performance numbers)

Now tested with each commit tested on Debian/Ubuntu.

`abigail` reports some "indirect sub-type change" of `PacketPipeline::comp_`. Actual testing using old Protonect with new libfreenect.so works. I consider this OK, unless any ABI experts point out the problem.

Interestingly there is a performance regression from kernel 3.19 (230Hz) to 4.2 (160Hz) without userspace change. The kernel performance regression is fixed in https://github.com/torvalds/linux/commit/f87a780f07b22b6dc4642dbaf44af65112076cb8 (4.4-rc7). For Ubuntu kernels, 4.2.0-28.33~14.04.1, or 4.4+ have this fixed. Kernels of 4.0 or earlier are not affected.


xlz:
@christiankerl I implemented your original idea of memory pool of size 2. As I expected, this is tricky to get right across multiple threads, as there is no explicit guarantee (there are some implicit conditions, but easy to get wrong) that the frame will be freed before the next allocation. I used your design such that a potential future improvement of zerocopy between processors and frame listeners can be done in similar ways without changing the API.

@rastaxe @tlind @guanchar The VAAPI feature is now ready for merge. Can you give it a run, and I'll merge it next week.

Note that there seems to be some performance regression between kernel 3.19 and 4.x.

xlz:
v2: Improved logging.

xlz:
The kernel performance regression is fixed in https://github.com/torvalds/linux/commit/f87a780f07b22b6dc4642dbaf44af65112076cb8 (4.4-rc7). For Ubuntu kernels, 4.2.0-28.33~14.04.1, or 4.4+ have this fixed. Kernels of 4.0 or earlier are not affected.

xlz:
v3: Added docs

xlz:
There is one more thing: libfreenect2 must be able to detect VAAPI hardware support at runtime and fall back to TurboJPEG, otherwise things will not work by default for non-VAAPI.

xlz:
v4: Let vaapi fallback to turbojpeg when not available at runtime

xlz:
v5: Added a note on platform and kernel compatibility. Print feature list during `cmake`

src/packet_pipeline.cpp
```diff
@@ -40,6 +40,13 @@ static RgbPacketProcessor *getDefaultRgbPacketProcessor()
 {
 #if defined(LIBFREENECT2_WITH_VT_SUPPORT)
   return new VTRgbPacketProcessor();
+#elif defined(LIBFREENECT2_WITH_VAAPI_SUPPORT)
+  RgbPacketProcessor *vaapi = new VaapiRgbPacketProcessor();
+  if (vaapi->good())
```
christiankerl: could you add a factory method to VaapiRgbPacketProcessor like tryCreate, which returns true if vaapi is available and false otherwise, depending on this you choose the code path here. this would avoid extension of the packetprocessor api with the good() method which is only used here (as far as I can see)?!

include/internal/libfreenect2/packet_processor.h
```diff
@@ -48,11 +138,32 @@ class PacketProcessor
...
 
+  virtual bool good() { return true; }
+
   /**
    * A new packet has arrived, process it.
    * @param packet Packet to process.
    */
   virtual void process(const PacketT &packet) = 0;
+
+  virtual Allocator *getAllocator() { return &default_allocator_; }
```
christiankerl: I see the point in adding this method, but shouldn't it be protected? users of packetprocessor only call allocateBuffer and releaseBuffer, right?

include/internal/libfreenect2/rgb_packet_processor.h
```diff
@@ -108,5 +109,22 @@ class VTRgbPacketProcessor : public RgbPacketProcessor
...
+
+class VaapiRgbPacketProcessor : public RgbPacketProcessor
+{
+public:
+  VaapiRgbPacketProcessor();
+  virtual ~VaapiRgbPacketProcessor();
+  virtual bool good();
+  virtual Allocator *getAllocator();
+protected:
+  virtual void process(const libfreenect2::RgbPacket &packet);
```
christiankerl: btw. why are all process methods of all rgb packet processor implementations protected, the process method is public in the PacketProcessor class

src/packet_pipeline.cpp
xlz: I could. I am thinking of adding check against errors in packet processors. Errors can happen anytime at runtime in packet processors. Once errors happen, there should be a way to report this to the user. Something outside of packet processors needs the check for the error state. I chose the interface as `PackteProcessor::good()`. What do you think?

include/internal/libfreenect2/async_packet_processor.h
```diff
@@ -119,6 +135,7 @@ class AsyncPacketProcessor : public PacketProcessor<PacketT>
       {
         // invoke process impl
         processor_->process(current_packet_);
+        releaseBuffer(current_packet_);
```
christiankerl: why do you call releaseBuffer here? shouldn't the stream parser do this? otherwise you assume that every packetprocessor is wrapped with an asyncpacketprocessor, right?

include/internal/libfreenect2/rgb_packet_processor.h
xlz: I don't know. I was just copying the definitions. Is there originally some intention to disallow directly access to the packet processor implementations outside of the async wrapper?

christiankerl: I think its a bug, because for the depthpacketprocessor its always public...

include/internal/libfreenect2/async_packet_processor.h
xlz: The stream parser passes the buffer asynchronously so it can not wait after `process()` finishes and free the buffer.

I have assumed, or found very packet processor is wrapped in the async one, yes, and reduced multiple copies of `releaseBuffer()` to just this one.

include/internal/libfreenect2/rgb_packet_processor.h
xlz: OK I'll fix this one.

include/internal/libfreenect2/packet_processor.h
xlz: I'll make it protected.

src/packet_pipeline.cpp
christiankerl: ok, didn't know this

include/internal/libfreenect2/async_packet_processor.h
christiankerl: ok, so basically every packetprocessor should call releaseBuffer at the end of process?

xlz: Yes. In theory releaseBuffer should be called as soon as the access to it is finished, but right now no new allocateBuffer will be called before `processor->ready()` becomes true, so putting it in the main loop of async processor is OK.

include/internal/libfreenect2/packet_processor.h
```diff
@@ -29,9 +29,99 @@
...
+
+template <class AllocatorT>
+class PoolAllocator: public Allocator
+{
+private:
+  AllocatorT allocator;
+  Buffer *buffers[2];
+  mutex locks[2];
+public:
+  PoolAllocator() : buffers() {}
```
christiankerl: maybe pass the base allocator as argument, store internally Allocator *allocator. afterwards, this one doesn't need to be templated anymore.

include/internal/libfreenect2/async_packet_processor.h
christiankerl: ok, maybe add a comment mentioning what we discussed here

xlz: Will do.

christiankerl:
did you try the allocator stuff with your cuda depth processor? just want to see if the api is sufficient or needs more changes.

include/internal/libfreenect2/packet_processor.h
xlz: Then each packet processor will need to define a default constructor to set the pointer. I was lazy so didn't do that.

Is there something bad to the template?

xlz:
I had not. I have been cleaning up the cuda series since yesterday.

include/internal/libfreenect2/packet_processor.h
christiankerl: currently this forces all allocators you want to use with poolallocator to have a ctor without arguments

xlz: True, I changed VaapiAllocator's constructor from having arguments to having no arguments (the arguments are later set using `PoolAllocator::baseAllocator().setVAContexts()`).

The alternatives are:
* Change the constructor of every packet processor to set the pointer - I feel like this is changing too many things.
* Set the pointer in `allocatorBuffer()` and `releaseBuffer()` - I find this ugly.

christiankerl: you could add a no argument ctor to poolallocator which sets the pointer to a newallocator

xlz: Good idea. Will do this.

xlz:
v6:

* Made `*RgbPacketProcess:process()` public from protected
* Changed PoolAllocator from a template to accepting base allocator in its constructor.
* Moved the allocator definitions to allocator.{cpp,h} replacing double_buffer.{cpp.h}

Major changes, need retesting.

xlz:
v7: Cosmetic

xlz:
I tried it with cuda processor. The API is OK.

xlz:
v8: Fixed a memory leak in `NewAllocator::free()`. Removed unnecessary `releaseBuffer()` in `VaapiRgbPacketProcessor::process()`.

xlz:
v9: Changed `PoolAllocator::allocate()` from `try_lock(); lock()` to a single `buffers[next].lock(); next = !next;`. Added comments specifying `PoolAllocator`'s behavior.

PoolAllocator can allocate in the same thread at most twice, then the next allocate() will block until free()'d by another thread (but this won't happen between stream parsers and processors - this blocking is intended for between processors and frame listeners).

PR #574, opened 2016-02-14
Delay start stream in OpenNI2 driver

hanyazou:
This fix delay calling Freenect2Device::start() and avoid calling close() which issue the shutdown USB command to the sensor device. You should not call Freenect2Device::close() every time OpenNI2 device closed because some OpenNI2 client call open/close repeatedly only to probe some properties of the device.


xlz:
I don't see a fixed sleep, so what do you mean by delay? A logical delay that put startStream after something else?

Also is this tested?

hanyazou:
Current implementation is calling start() after freenect2.openDevice() immediately. With this fix, Freenect2Device::start() will be called from openni::VideoStream::start(). If openni::VideoStream will be closed without start(), the driver does not call Freenect2Device::close(). (This might introduce some memory leak. But it is better than hangup.)

I've tested this with NiViewer and pcl_openni2_viewer on my mac.

PR #579, opened 2016-02-18
Opencl improvements

wiedemeyer:
Which this PR includes:
- Custom Allocator and Frames for OpenCL using pinned memory
- Removed arrays for tables, instead data is directly written to the OpenCL buffers
- OpenCL buffers are now created only once on initialization
- Added definition to enable profiling for the OpenCL kernels
- Minor improvements for the first stage

With this improvements the OpenCL depth processor takes between 1.09 ms and 1.11 ms on my system, instead of ~2 ms. Most due to the pinned memory.

### DIFF
--- a/include/internal/libfreenect2/depth_packet_processor.h
+++ b/include/internal/libfreenect2/depth_packet_processor.h
@@ -175,7 +175,9 @@ public:
   virtual void loadXZTables(const float *xtable, const float *ztable);
   virtual void loadLookupTable(const short *lut);
 
+  virtual bool good();
   virtual const char *name() { return "OpenCL"; }
+
   virtual void process(const DepthPacket &packet);
 protected:
   virtual Allocator *getAllocator();
--- a/include/internal/libfreenect2/depth_packet_processor.h
+++ b/include/internal/libfreenect2/depth_packet_processor.h
@@ -177,6 +177,8 @@ public:
 
   virtual const char *name() { return "OpenCL"; }
   virtual void process(const DepthPacket &packet);
+protected:
+  virtual Allocator *getAllocator();
 private:
   OpenCLDepthPacketProcessorImpl *impl_;
 };
--- a/src/opencl_depth_packet_processor.cl
+++ b/src/opencl_depth_packet_processor.cl
@@ -24,8 +24,7 @@
  * either License.
  */
 
-#define PHASE_SIN (float3)(PHASE_IN_RAD0_SIN, PHASE_IN_RAD1_SIN, PHASE_IN_RAD2_SIN)
-#define PHASE_COS (float3)(PHASE_IN_RAD0_COS, PHASE_IN_RAD1_COS, PHASE_IN_RAD2_COS)
+#define PHASE (float3)(PHASE_IN_RAD0, PHASE_IN_RAD1, PHASE_IN_RAD2)
 #define AB_MULTIPLIER_PER_FRQ (float3)(AB_MULTIPLIER_PER_FRQ0, AB_MULTIPLIER_PER_FRQ1, AB_MULTIPLIER_PER_FRQ2)
 
 /*******************************************************************************
@@ -47,8 +46,8 @@ float decodePixelMeasurement(global const ushort *data, global const short *lut1
   return (float)lut11to16[(x < 1 || 510 < x || col_idx > 352) ? 0 : ((data[data_idx0] >> upper_bytes) | (data[data_idx1] << lower_bytes)) & 2047];
 }
 
-void kernel processPixelStage1(global const short *lut11to16, global const float *z_table, global const float3 *p0_sin_table, global const float3 *p0_cos_table,
-                               global const ushort *data, global float3 *a_out, global float3 *b_out, global float3 *n_out, global float *ir_out)
+void kernel processPixelStage1(global const short *lut11to16, global const float *z_table, global const float3 *p0_table, global const ushort *data,
+                               global float3 *a_out, global float3 *b_out, global float3 *n_out, global float *ir_out)
 {
   const uint i = get_global_id(0);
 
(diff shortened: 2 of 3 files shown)


wiedemeyer:
Having two tables instead of one is actually slower than to compute them directly. We talked about it here: #222. Or did you meant something else?

wiedemeyer:
I will update the PR with the suggested changes tomorrow.

xlz:
There are two other things. One is `PacketProcessor::good()` is now the API to propagate errors within packet processors. The specific semantics are not specified yet, but generally when `good()` returns false, the processor will not be touch anymore.

The other one is please use `opencl: Verb some short description` as commit subjects then put details in the body.

xlz:
When you're done with the PR, please also provide some benchmark results using steps in https://github.com/OpenKinect/libfreenect2/wiki/Performance

Results so far:


Probably all of it has to be redone after your OpenCL optimization.

wiedemeyer:
Here are the benchmark results.

<table>
<tr><th>Configuration</th><th>Depth (min, 5%, median, 95%, max, mean, std)</th><th>RGB (min, 5%, median, 95%, max, mean, std)</th><th>Thread per core usage</th></tr>
<tr><th colspan=4>Feb 19, 2016: Intel i7-4770K (@4.1GHz), GTX 980Ti; Ubuntu 14.04, kernel 4.2.0-29, gcc 4.8.5</th></tr>
<tr><td>CPU/TurboJPEG</td><td>194.328 196.617 200.837 212.289 225.055 mean=201.911 std=4.65475</td><td>12.0173 12.2656 13.1827 19.4198 22.4747 mean=13.6461 std=1.78962</td><td>CPU:90% TurboJPEG:40% USB:5% Reg:3%</td></tr>
<tr><td>OpenGL/TurboJPEG</td><td>3.22797 3.36801 8.02571 9.06995 108.705 mean=7.09752 std=2.96411</td><td>11.9735 12.2769 13.5689 19.4342 28.3156 mean=14.3881 std=2.23828</td><td>OpenGL:26% TurboJPEG:44% USB:6% Reg:20%</td></tr>
<tr><td>OpenCL/VAAPI</td><td>1.07144 1.08136 1.0924 1.145 2.46014 mean=1.1035 std=0.0599953</td><td>4.14765 4.1658 4.66865 7.72171 11.1335 mean=4.98485 std=1.18519</td><td>OpenCL:3% VAAPI:2% USB:6% Reg:18%</td></tr>
<tr><td>CUDA/VAAPI</td><td>0.857415 0.861542 0.868286 0.924719 3.31855 mean=0.882014 std=0.0699696</td><td>4.12401 4.14701 4.6825 10.9971 11.2794 mean=5.18491 std=1.60745</td><td>CUDA:5% VAAPI:2% USB:6% Reg:22%</td></tr>
</table>

wiedemeyer:
I added all the recommended changes except the `p0_table` array as a class member, I changed it to dynamically allocated memory, because `p0_table` is just used once and then never again.
The rebase of the fork to the upstream master seems to have removed the comments on the commits, sorry for that.

xlz:
Which GPU (Intel or Nvidia) did the OpenCL case use?

wiedemeyer:
Nvidia, i edited the previous post.

xlz:
I reviewed some of the changes. There are a lot of back and forth so it's no longer practical to reorganize this into standalone functional patches. I'll merge the code as-is but will format the subjects and commit messages.

Do you want to add something in CMakeLists.txt to turn on LIBFREENECT2_WITH_PROFILING_CL more easily or revert it to a file-level macro like `#ifdef PROFILE_CL` which can only be enabled by developer editing CMakelists.txt? I suppose you didn't turn it on during the benchmarking?

xlz:
I have added your results to the wiki (Do you have permissions to edit it?)

Also after reviewing your patches I realize the CUDA processor still lacks some of the improvement you added here, good catches.

wiedemeyer:
The back and forth wasn't intended. 
We could add setting the definition to the CMakeLists, but I don't think it is important, because nearly nobody will use it. I just wanted the code to be there in case of further improvements. I am ok with putting a `#define LIBFREENECT2_WITH_PROFILING_CL` (or however we name it) in the source code to enable profiling. And yes I disabled the CL profiling, when profiling the processors. 
I think I can edit the wiki, there is an edit button, but I haven't tried to change anything yet. I can also benchmark some other combinations next week if needed. It would also be great, if the profiling would measure the CPU utilization of the threads itself, rather then trying to find a good average from the top output.

xlz:
You can add the benchmark results directly to the wiki next time.

xlz:
Well, per core usage is not that important. Adding the stats is quite intrusive so I give up.

xlz:
Manually merged.

If you don't feel like it, just ignore and not collect the per core usage. So far it is only useful in identifying the memory access on Tegra is extremely slow.

xlz:
@wiedemeyer Can you run the CUDA test again? Hopefully VAAPI will have smaller variance.

wiedemeyer:
yes, I will run it this afternoon.

wiedemeyer:
@xlz 
I profiled all combinations and put the results into the wiki. I also created the gnuplot image, but I don't know where and how to replace the current one.

But there is something wrong with the CUDA processor. When I run VAAPI with CUDA Protonect uses ~70% CPU, that is much compared to OpenCL with only ~33% CPU, or OpenGL with ~41% CPU. The CPU usage refers to the usage of the whole process as show by htop. You can also see that in the results from the profiling in the wiki.

xlz:
I'll render the image.

According the your usage data, only the registration in the main thread has unusual CPU usage. It's weird. There really is nothing special in the registration routines. I'll take a look at what's going on on my side. You can check too.

wiedemeyer:
I ran kinect2_bidge and subscribed to `/kinect2/sd/camera_info`. This way the kinect2_bridge does nearly nothing, except starting capturing and publishing the camera_info. There it made no difference between using OpenCL or Cuda. The CPU usage of the whole process was just 12%. Why would Protonect without a viewer take so much more resources, can it be the profiling? I don't have much time now to look into it, but maybe later.

xlz:
The slowness in Registration is caused by a wrong write combined flag used in CUDA. Fixed in https://github.com/OpenKinect/libfreenect2/commit/18d1cff311703da730680ee5f8b025f70a2fff43

wiedemeyer:
But it fixed the high CPU load with CUDA. I updated the wiki performance entries.

PR #585, opened 2016-02-22
Add -frames option to the Protonect

hanyazou:
this option is convenient for test automation


hanyazou:
tested on Mac and Linux (Jetson TK1)

xlz:
Manually merged.

I changed int64_t to size_t, and strtoll to strtol (strtoll is C++11).

hanyazou:
@xlz Thank you. But why don't you use the second argument of strtol()? You can't specify zero for the frame count with your change. And strtol may ignore trailing garbage like '-frames 1cl' ('cl' is just ignored). I'm just curious while it might be okey.

xlz:
I didn't think `-frames 0` has any purpose. `-frames 1` can do stress test already.

I was going to use `atoi`, though. I didn't think this needs super strict checking.

Issue #590, opened 2016-02-24
libfreenect2 builds but doesn't run Protonect


robozo:
Hi, I have a modified Protonect running just fine on a mid-2012 macbook Pro running 10.8.5,
and am attempting to install libfreenect2 and Protonect on a nearly identical machine, and am having
problems.  Things have changed with the new version I guess.

I got a libfreenect2 build error kVTVideoDecoderSpecification_EnableHardwareAcceleratedVideoDecoder not defined in
vt_rgb_packet_processor.cpp.  I guess this is new in 10.9, but useless, there is no hardware
jpeg decoding on mac, is there?

So I manually defined that identifier = 0 in vt_rgb_packet_processor.cpp and libfreenect2 built,
but Protonect won't run.  It finds the valid Kinect, with serial number ####, then bombs out
with Trace/BPT trap: 5.  Google tells me this is a dynamic library problem, but I don't know
how to proceed, and whether this is related to my earlier problem.

Should I try reverting to an earlier library?

Any suggestions welcome!

rob

...for the moment gave up trying to get new version to work.  copied libfreenect2.dylib from
old machine into /usr/local/lib in the new machine, things are working.

xlz:
@fran6co This time we have 10.8.

https://developer.apple.com/library/mac/releasenotes/General/APIDiffsMacOSX10_9/VideoToolbox.html
```
Added kVTVideoDecoderSpecification_EnableHardwareAcceleratedVideoDecoder
Added kVTVideoDecoderSpecification_RequireHardwareAcceleratedVideoDecoder
```
What happens if you use kVTVideoDecoderSpecification_RequireHardwareAcceleratedVideoDecoder?

fran6co:
I think I can remove the hardware accelerated stuff because it's not actually accelerated in Mac... It is for IOS but it should be using it by default. I'll create the PR and test it for 10.8

fran6co:
Just tried in 10.8 and works no problem.

robozo:
Thanks!  Sorry about 10.8, is stable for me for a lot of old code.
It could have been worse, could have been classic mac 9.1!

fran6co:
The patch was trivial, as long Apple doesn't add JPEG hardware acceleration by setting some flags we are going to be ok.

PR #592, opened 2016-02-24
10.8 compatibility

fran6co:
Fixes https://github.com/OpenKinect/libfreenect2/issues/590


xlz:
Then is VT faster than TurboJPEG?

fran6co:
In my test and according to @RyanGordon it is. Not by much because it's still CPU based. If Apple decides to enable hardware acceleration (because the hardware supports Intel's Quick Sync) it's going to be through this API.

xlz:
Maybe try benchmarking? https://github.com/OpenKinect/libfreenect2/wiki/Performance

fran6co:
I will when I get some time to do it.

Issue #539, opened 2016-01-24
The device disappear after calling close() for a few seconds.


hanyazou:
While investigating #517, I've noticed that it takes a few seconds to restart the Kinect device after calling close(). And the USB deice disappear for a second. This means that you must wait for a while before you call enumerateDevices() and openDevice() after calling close(). This issue might depend Mac OS X.

You can reproduce this issue with stock Protonect.

case 1 (hungup)
-------
> $ ./bin/Protonect; sleep 0; ./bin/Protonect

1. Stop with ESC key after the first Protonect shows the image.
2. Second Protonect does not show any images.

> [Info] [Freenect2DeviceImpl] enabling usb transfer submission...
[Info] [Freenect2DeviceImpl] submitting usb transfers...
[Info] [Freenect2DeviceImpl] started
device serial: 005697142447
device firmware: 4.0.3912.0
[Debug] [DepthPacketStreamParser] not all subsequences received 1022
// no messages. it seems that no data come from the device.

In this case Ctrl-C and ESC can not stop the second Protonect. It requires kill -KILL.

case 2 (error)
-------
> $ ./bin/Protonect; sleep 1; ./bin/Protonect

1. Stop with ESC key after the first Protonect shows the image.
2. Second Protonect can't detect the device.

>[Info] [Freenect2Impl] enumerating devices...
[Info] [Freenect2Impl] 12 usb devices connected
[Info] [Freenect2Impl] found 0 devices
no device connected!

case 3 (this works fine)
-------
> $ ./bin/Protonect; sleep 3; ./bin/Protonect

1. Stop with ESC key after the first Protonect shows the image.
2. Second Protonect shows the image also.

xlz:
> The device disappear after calling close() for a few seconds.

What is wrong with this?

hanyazou:
@xlz Please refer case 1 and case 2. It fails randomly depending on the timing. This type of behavior debase stability of system which uses this library.

For example, this test script fails randomly because of this issue.

>#!/bin/bash
test()
{
  $* &
  pid=$!
  sleep 5
  kill $pid || exit
  sleep 1  # longer wait required
}
test UserViewer
test Protonect
test NiViewer

xlz:
The Kinect reboots and it takes 3 seconds. What can you do about it?

Case 1 fails because you try to use the Kinect when it is shutting down. Case 2 fails because the Kinect is still booting.

How to not shut down the Kinect when the program exits?

[...]

PR #597, opened 2016-02-28
Reboot device

hanyazou:
Avoid being stuck in the status checking describing in
https://github.com/OpenKinect/libfreenect2/issues/539#issuecomment-189577849.

I tested that Protonect ran normally and reboot when timeout occurred in startStreams() on Mac OS X 10.11 and Ubuntu 14.04 (Jetson TK1) with these cmake configurations,
-DENABLE_CXX11=ON/OFF
-DENBELE_OPENGL=ON/OFF

### DIFF
--- a/include/internal/libfreenect2/protocol/command_transaction.h
+++ b/include/internal/libfreenect2/protocol/command_transaction.h
@@ -47,6 +47,7 @@ public:
   CommandTransaction(libusb_device_handle *handle, int inbound_endpoint, int outbound_endpoint);
   ~CommandTransaction();
 
+  void setHandle(libusb_device_handle *handle);
   bool execute(const CommandBase& command, Result& result);
 private:
   libusb_device_handle *handle_;
--- a/include/internal/libfreenect2/protocol/usb_control.h
+++ b/include/internal/libfreenect2/protocol/usb_control.h
@@ -79,6 +79,7 @@ public:
     Error
   };
 
+  void setHandle(libusb_device_handle *handle);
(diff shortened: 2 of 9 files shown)


xlz:
I don't know what bug this PR is trying to fix. There is no bug report specific to this PR, and yes you have said something in #539:

> Sometimes the Protonect stopped around status 0x090000: 1024 until I did un-plug and re-plug the device manually.

I have tested on 5 different kinds of machines and multiple operating systems and never one did I see this behavior. No other Mac users have reported this. Maybe it's a problem with your own hardware. Maybe it's a rare enough theoretical problem. You have to prove how big a problem this it.

I have had a simple suggestion:

> Maybe you can set a limit of retries for that loop and see what happens.

Which means adding a counter here https://github.com/OpenKinect/libfreenect2/blob/master/src/libfreenect2.cpp#L716
```cpp
for (uint32_t status = 0, last = 0, i = 0; (status & 1) == 0 && i < 1000; last = status, i++)
```
There was no `ReadStatus0x090000Command` at all previously, and it still worked. If this `ReadStatus0x090000Command` somehow creates a problem, you can simply skip it.

I object to the "hammering the API when it goes wrong until it works" approach. It doesn't work like that, unless you have good reasons.

xlz:
The first two commits mysteriously shift around the APIs without justifications. The third commit does the API hammering thing.

I think I have said this. That is heave-handed. It might be acceptable if it is limited to very specific scenarios, and it can be verified by other Mac users, and it solves a real problem instead of a theoretical possibility.

Whatever the bug this PR tries to fix, you have to try again.

PR #646, opened 2016-04-25
Implement non-default pipeline selection from OpenNI2 side.

smokhov:
This PM implements a way to select non-default pipeline when using OpenNI2 driver with NiTE samples. This is done via an environment variable `LIBFREENECT2_NITE_PIPELINE_TYPE` that takes values `GL`, `CL`,  `CUDA`, and `CPU`. If none are selected, the default pipeline is selected as before. The change can be done at run-time. I needed that feature being stuck for a while with default TurboJPEG and GL stuff. Please consider for inclusion.

### DIFF
--- a/src/openni2/DeviceDriver.cpp
+++ b/src/openni2/DeviceDriver.cpp
@@ -500,7 +500,9 @@ namespace Freenect2Driver
             WriteMessage("Opening device " + std::string(uri));
             int id = uri_to_devid(iter->first.uri);
             DeviceImpl* device = new DeviceImpl(id);
-            device->setFreenect2Device(freenect2.openDevice(id)); // XXX, detault pipeline // const PacketPipeline *factory);
+            // The LIBFREENECT2_PIPELINE variable allows to select
+            // the non-default pipeline
+            device->setFreenect2Device(freenect2.openDevice(id));
             device->setConfigStrings(config);
             iter->second = device;
             return device;
--- a/src/openni2/DeviceDriver.cpp
+++ b/src/openni2/DeviceDriver.cpp
@@ -500,7 +500,43 @@ namespace Freenect2Driver
             WriteMessage("Opening device " + std::string(uri));
             int id = uri_to_devid(iter->first.uri);
             DeviceImpl* device = new DeviceImpl(id);
-            device->setFreenect2Device(freenect2.openDevice(id)); // XXX, detault pipeline // const PacketPipeline *factory);
+
+			const char *pipeline_type_env;
+			// GL, CL, CUDA, CPU
+			pipeline_type_env = std::getenv("LIBFREENECT2_NITE_PIPELINE_TYPE");
+			std::string pipeline_type;
+
+			if (pipeline_type_env)
+			{
+				pipeline_type = std::string(pipeline_type_env);
+			}
+
+			libfreenect2::PacketPipeline* pipeline = NULL;
+
+#if defined(LIBFREENECT2_WITH_OPENGL_SUPPORT)
+			if (pipeline_type == "GL")
+				pipeline = new libfreenect2::OpenGLPacketPipeline();
+#elif defined(LIBFREENECT2_WITH_CUDA_SUPPORT)
+			if (pipeline_type == "CUDA")
+				pipeline = new libfreenect2::CudaPacketPipeline();
+#elif defined(LIBFREENECT2_WITH_OPENCL_SUPPORT)
+			if (pipeline_type == "CL")
+				pipeline = new libfreenect2::OpenCLPacketPipeline();
+#else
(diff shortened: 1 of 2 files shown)


src/openni2/DeviceDriver.cpp
```diff
@@ -500,7 +500,43 @@ namespace Freenect2Driver
             WriteMessage("Opening device " + std::string(uri));
```
xlz: Do you need `#include <libfreenect2/packet_pipeline.h>`? (Just asking, I don't know.)

src/openni2/DeviceDriver.cpp
```diff
@@ -500,7 +500,43 @@ namespace Freenect2Driver
...
             int id = uri_to_devid(iter->first.uri);
             DeviceImpl* device = new DeviceImpl(id);
-            device->setFreenect2Device(freenect2.openDevice(id)); // XXX, detault pipeline // const PacketPipeline *factory);
+
+			const char *pipeline_type_env;
+			// GL, CL, CUDA, CPU
+			pipeline_type_env = std::getenv("LIBFREENECT2_NITE_PIPELINE_TYPE");
```
xlz: I think `LIBFREENECT2_OPENNI2_PIPELINE` is more appropriate.

src/openni2/DeviceDriver.cpp
```diff
@@ -500,7 +500,43 @@ namespace Freenect2Driver
             WriteMessage("Opening device " + std::string(uri));
             int id = uri_to_devid(iter->first.uri);
             DeviceImpl* device = new DeviceImpl(id);
-            device->setFreenect2Device(freenect2.openDevice(id)); // XXX, detault pipeline // const PacketPipeline *factory);
+
+			const char *pipeline_type_env;
+			// GL, CL, CUDA, CPU
```
xlz: Please clean up the indentation. Replace tabs with spaces.

src/openni2/DeviceDriver.cpp
```diff
@@ -500,7 +500,43 @@ namespace Freenect2Driver
...
+			{
+				pipeline_type = std::string(pipeline_type_env);
+			}
+
+			libfreenect2::PacketPipeline* pipeline = NULL;
+
+#if defined(LIBFREENECT2_WITH_OPENGL_SUPPORT)
+			if (pipeline_type == "GL")
+				pipeline = new libfreenect2::OpenGLPacketPipeline();
+#elif defined(LIBFREENECT2_WITH_CUDA_SUPPORT)
```
xlz: I think you should not use #elif here. Try this
```cpp
#if defined(...OPENGL...)
   ...
#endif
#if defined(...CUDA...)
   ...
#endif
...
```

xlz:
The idea is ok. Please look at the comments. Fix them and report the test results then I'll merge it.

src/openni2/DeviceDriver.cpp
smokhov: Not necessary, it is already included via `#include <libfreenect2/libfreenect2.hpp>`, so it compiled as is.

src/openni2/DeviceDriver.cpp
smokhov: You are right; this would be a bug -- it worked for me, but I had OpenCL pre-set so it did not show up and I can't test CUDA. I will fix it.

smokhov:
Thanks, I will amend tabs to spaces and rename the variable. I guess it is useful also to let people know in the REAMDE of it, so I'll try to fix it all up in the next couple of commits.

xlz:
Can you add a wider scoped variable `LIBFREENECT2_PIPELINE` to https://github.com/OpenKinect/libfreenect2/blob/master/src/libfreenect2.cpp#L901 so users other than openni2 can also specify the pipeline without changing any code?

smokhov:
Sure, I can change it to `LIBFREENECT2_PIPELINE`; in my original I was just being conservative and minimalistic.

xlz:
It may be confusing because createDefaultPacketPipeline() is indeed using #elif, but that is because it is selecting one pipeline at build time so it can do that. Now what you need is selecting at runtime so you can't use elif.

There are several environment variables I intend to add to the docs, but you happen to have this straightforward idea at the last minute before 0.2 release so I'll wait.

smokhov:
Yes, I understand; I'll get on it; it makes sense.

smokhov:
@xlz -- please check recent commits; should not be something too extraordinary. Am I correct in assuming you would document `LIBFREENECT2_PIPELINE` yourself? :)

xlz:
Manually merged.

I changed and split a function out of the code for future improvement.

hanyazou:
merged in 7b83c7f90e19cccccdfd55d643f72c310070d6cb

PR #655, opened 2016-05-12
Avoid overriding the PKG_CONFIG_PATH environment variable

agilmor:
A very small change to append the local path to the current PKG_CONFIG_PATH on the system.
Without this change, if libusb is on the system but not in the default paths, cmake won't find it even if PKG_CONFIG_PATH is correctly set.

I've only tested it on cmake 3.0, and maybe [this improvement on cmake 3.1](http://public.kitware.com/Bug/view.php?id=12926) solves the overriding...?
But anyway, append paths to the system's PKG_CONFIG_PATH seems better than overrinding it.


xlz:
Looks ok.

PR #690, opened 2016-07-19
ubuntu 16.04

geoffviola:
Added some support for the Ubuntu 16.04's cuda and gcc. There's probably a better to get compatibility with cuda than to copy the helper_math.h from cuda 5.


andre-nguyen:
This partially worked for me but I also had to add "-D_MWAITXINTRIN_H_INCLUDED" to get it building on 16.04.

xlz:
> There's probably a better to get compatibility with cuda than to copy the helper_math.h from cuda 5.

What do you mean? `<helper_math.h>` is sourcing the header from the CUDA toolkit samples directory you are using. Not CUDA 5?

Anyway it's not possible to copy Nvidia's code under our license.

I will fix 16.04 build issues separately.

geoffviola:
>>There's probably a better to get compatibility with cuda than to copy the helper_math.h from cuda 5.

>What do you mean? <helper_math.h> is sourcing the header from the CUDA toolkit samples directory you are using. Not CUDA 5?

I copied the file from an older CUDA API into my commit just to get it to compile. CUDA 5 does not have that header file.

> Anyway it's not possible to copy Nvidia's code under our license.

Not using that header file is a good, long-term solution

xlz:
Also, gcc later than 4.9 is not supported by CUDA 7.5 anymore. So use CUDA 8 instead.

geoffviola:
>  Also, gcc later than 4.9 is not supported by CUDA 7.5 anymore. So use CUDA 8 instead.

Oh, yeah I was on Ubuntu 16.04. So GCC and CUDA would have been 5.4 and 7.5 respectively. The header file was from CUDA 5.

PR #701, opened 2016-08-01
Added gstjpeg sources also for L4T 24.1

marketto89:


xionglingfeng:
This fix is not enough for R24.1.

    FIND_LIBRARY(TegraJPEG_LIBRARIES

needs to add `/usr/lib/aarch64-linux-gnu/tegra`

xlz:
Should be fixed in 07a7424cc2e22989942818fc74f6d940a4fa6b30.

PR #716, opened 2016-09-02
Optimize OpenGL depth packet processor image flipping

SirDifferential:
This pull request optimizes OpenGL depth packet processor by reducing the CPU time of downloadToNewFrame(). In the old system textures arrive from the graphics card upside down and they are flipped on the CPU. This pull request adds some tweaks that skip the flipping on CPU and instead parses the texture coordinates as flipped in the first place.

[Very Sleepy profiler output before optimization](http://plantmonster.net/koodailut/libfreenect2/freenect_optimize001.png)
[Very Sleepy profiler output after optimization](http://plantmonster.net/koodailut/libfreenect2/freenect_optimize002.png)

If there are issues with this PR, please let me know.

### DIFF
--- a/CONTRIB
+++ b/CONTRIB
@@ -1,11 +1,34 @@
-For a list of libfreenect2 authors, please see the Git history of the project,
-such as with 'git log'. This file will contain a list of authors on release
-snapshots of libfreenect2.
-
-THIS FILE IS NOT VALID FOR NON-GIT DISTRIBUTIONS. If you wish to redistribute
-libfreenect source code outside of the Git infrastructure, you MUST either use
-a release version or generate a complete CONTRIB file valid for the source Git
-revision tree that you use for your distribution. Redistributions of libfreenect
-source code without a valid CONTRIB file and without Git metadata are not
-compliant with the license.
-
+Albert Hofkamp <a.t.hofkamp@tue.nl>
+Alistair <alistair.wk@gmail.com>
+augmenta <augmenta@augmenta.home>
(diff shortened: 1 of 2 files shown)


src/shader/stage1.fs
```diff
@@ -94,13 +94,14 @@ vec2 processMeasurementTriple(in ivec2 uv, in usampler2DRect p0table, in int off
 void main(void)
 {
   ivec2 uv = ivec2(TexCoord.x, TexCoord.y);
+  ivec2 uv_flip = ivec2(TexCoord.x, 423 - TexCoord.y);
     
   bool valid_pixel = 0.0 < texelFetch(ZTable, uv).x;
```
christiankerl: uv_flip missing here?

SirDifferential:
Chris: Yes, I think so. There's also a couple in stage2.fs I did some initial tests and spent a while reading what the Z and X tables do, and it indeed seems like they need to be read using the flipped coordinates as well.

However, I also noticed an odd black region in the image that appears every 30 frames or so. I'd say hold this PR for now until I can investigate more.

SirDifferential:
Er, I accidentally pushed my Intel media SDK changed in the master branch. I'll see if I can fix this somehow to remove them from this PR.

xlz:
While you're at it, can you make it an option to flip the image so existing users would not get unexpected results? You can leave the option variable internal, or setting it from an environment variable, so we can expose this option later through the API.

PR #760, opened 2016-11-30
implemented VideoStream::convertDepthToColorCoordinates (openni2)

maxime-tournier:
Hello,

I stumbled upon an unimplemented method in openni2 driver when trying to convert between depth/color coordinates: `VideoStream::convertDepthToColorCoordinates`. So here is a fix.

The PR is pretty straightforward: 
- exposed `libfreenect2::Registration::apply` in `Freenect2Driver::Registration`, 
- adapted what's needed in `VideoStream::convertDepthToColorCoordinates`.

It seems working for my needs and I have NiTE2 skeleton tracking on top of color stream (yay!), but since this is my first time hacking on `libfreenect2` please let me know if I did something silly.

Best regards,

### DIFF
--- a/src/openni2/Registration.cpp
+++ b/src/openni2/Registration.cpp
@@ -46,7 +46,8 @@ void Registration::depthFrame(libfreenect2::Frame* frame) {
 }
 
 
-static libfreenect2::Registration* make_registration(libfreenect2::Freenect2Device* dev) {
+static libfreenect2::Registration* make_registration(libfreenect2::Freenect2Device* dev)
+{
   libfreenect2::Freenect2Device::ColorCameraParams colCamParams = dev->getColorCameraParams();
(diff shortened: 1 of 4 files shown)


xlz:
Fix indentation.

Otherwise @hanyazou

xlz:
This overwrite of `reg` looks to me very much like memory leak. I don't even know how it works in the openni2 part or it is even correct code. You have to come up with a usage example for testing for me to merge it.

maxime-tournier:
I am not sure what you mean by "overwrite".

I just factored out the existing registration creation code in a static function `make_registration`, then call the registration creation the exact same way in the two methods (`Freenect2Driver::Registration::colorFrameRGB888` and `Freenect2Driver::Registration::depthToColor`).

More precisely: I first check for null `reg` member then call the registration creation accordingly. Again, this is the exact same code as originally.

Am I missing something?

xlz:
I get it now.

Issue #749, opened 2016-11-03
OpenNI2 driver does not free devices before process exits


SirDifferential:
Overview Description:

When using the OpenNI2 driver on Windows or OS X, calling `openni::Device::open()` will claim the USB device and make it unusable by other processes.

Version, Platform, and Hardware Bug Found:

1. My own fork from latest master `5819b69` with a few changes to device enumeration and selecting the pipeline in DeviceDriver.cpp
2. `uname -a`: Darwin DelicodeMBP 15.5.0 Darwin Kernel Version 15.5.0: Tue Apr 19 18:36:36 PDT 2016; root:xnu-3248.50.21~8/RELEASE_X86_64 x86_64 and Windows 8.1 64-bit
3. `lsusb -t`: Not available on the platforms
4. `lspci -nn`: Not available on the platforms

Steps to Reproduce:

* Enumerate all devices by calling `openni::OpenNI::initialize()` followed by `openni::OpenNI::enumerateDevices()`
* Iterate through the found devices and open devices by uri: `device.open(dev_info.getUri())`, then close them by `device.close()`.
* Keep the process running
* In another process, perform the same steps

Actual Results:

In the second process:
    * On Windows, 0 devices are found
    * On OS X, a crash happens in libfreenect2.cpp at `Freenect2Device *Freenect2::openDevice(int idx, const PacketPipeline *pipeline, bool attempting_reset)` when the device is being reseted.

Expected Results:

Calling openni::Device::stop() is expected to free the USB device and make it available for other processes, but this does not happen.

Reproducibility:

Happens on Windows 8.1 and OS X 10.11 every time.

Additional Information:

The temporary hack I created for this is by changing device enumeration so that OpenNI2 only ever gets one device from the enumeration, even if multiple are connected:

* `libfreenect2.cpp::enumerateDevices()` reads the currently used serial numbers from an env var I set before calling `openni::OpenNI::initialize()`
* It adds only the first free sensor to the enumerated devices container, making sure only one sensor is ever enumerated.

This hack avoids the problem. I'm testing with 2 Kinects. Using libfreenect2 directly without OpenNI2 works.

I'm assuming the problem is related to the `DeviceDriver.cpp` function `DeviceImpl::stop()` which does not correctly free some OpenNI2 resources before the process exits.

hanyazou:
Could you please share your code or patch here?

SirDifferential:
https://github.com/SirDifferential/libfreenect2_openni2_tester

[...]

PR #762, opened 2016-12-01
Add serial to uri, fix for loop variable re-use

SirDifferential:
As per issue #749, it was agreed that serial numbers can be added in device URIs. This commit adds that change.

This commit also fixes a for loop re-using the `i` variable, causing some initializations to return wrong modes or wrong number of devices.


xlz:
I edited your code. The URI format in your code is incorrect.

PR #815, opened 2017-03-23
typo preventing correct display of markdown

eric-schleicher:
The ### linux section trapped inside \`\`\` block


xlz:
The fix is incomplete. Github switched to CommonMark days ago.

I'll clean it up. https://github.github.com/gfm/

Issue #744, opened 2016-10-24
Is it worth adding CUDA support for generating point cloud from RGB&Depth data


fengjim:
Overview Description:

I'm using libfreenect2 to collect RGB and Depth Data from Kinect2 devices on Linux (Ubuntu 14.04) and generate PCL point cloud based on that. 

The steps are generally: 1) libfreenect2::SyncMultiFrameListener::waitForNewFrame()  to get the RGB and Depth frames, 2) libfreenect2::Registration::apply() to align them, 3) loop through 512x424 matrix and call libfreenect2::Registration::getPointXYZRGB() to fill all the matrix elements.

According to the performance testing result, step 3# was the one taking most of time in the whole pipeline. I was thinking to use parallel programming (either CUDA with GPU or multiple thread on CPU) in Step 3# to improve the efficiency. However, considering libfreenect2 has already providing CUDA/OpenGL options of pipelines etc. , it might looks helpful if libfreenect2 could provide one more function along aside with getPointXYZRGB() using CUAD etc. to generate all the points, i.e. adding libfreenect2::Registration::getPointXYZRGB(const Frame* undistorted, const Frame* registered, float** depth, uint_8** color), where 'depth' point to a 3x512x424 array of float representing point (X, Y, Z) matrix and 'color' point to a 3x512x424 array of uinit_8 representing  color for related point. 

Would you please kindly share your comments/thoughts about this?

Thanks in advance!

xlz:
It's worth trying. 

You probably want to create a `class CudaRegistration` in `cuda_registration.cpp` and `registration.h`.

To interoperate with PCL, you'll need to look at the byte format of PCL. It looks like a pointcloud in PCL stores data in `vector<pcl::PointXYZRGB>` or `vector<pcl::PointXYZ>`. `pcl::PointXYZ` is 4 bytes. `pcl::PointXYZRGB` is 8 bytes.

fengjim:
@xlz, thanks a lot for your reply and suggestions. 

I will have a try then:)

hanshammel1337:
@fengjim Did you make any progress on this topic? I am also interested in it.

fengjim:
Hi @hanshammel1337 , umm, I haven't done any real progress other than checking GPU related materials.

You may go ahead to start it over:) would be appreciated that you can share your branch later after you start working on it.

blackzafiro:
[...]

PR #822, opened 2017-04-04
Cuda registration

blackzafiro:
I see there are a few things which can be improved, but this is a functional prototipe.  Please let me know what you think.


xlz:
> 	Will need thrust to create 3D clouds

Why?

CMakeLists.txt
```diff
@@ -125,6 +125,7 @@ SET(SOURCES
   include/libfreenect2/packet_pipeline.h
   include/internal/libfreenect2/packet_processor.h
   include/libfreenect2/registration.h
+  include/libfreenect2/cuda_registration.h
```
xlz: The API specific implementation class should be put in `registration.h`, not `cuda_registration.h`. See how CUDA depth processor doesn't have its own header instead its definition is in `depth_packet_processor.h`.

Here was the API design I outlined https://github.com/OpenKinect/libfreenect2/issues/744#issuecomment-255842615

include/libfreenect2/cuda_registration.h
```diff
@@ -0,0 +1,162 @@
...
+
+typedef thrust::tuple<float, float, float, float> TupleXYZRGB;
+
+/**
+ * Frame whose data is allocated on device.
+ */
+class LIBFREENECT2_API CudaDeviceFrame: public Frame
+{
+public:
+	/** Construct a new frame.
```
xlz: You should fix your indentation. This project uses 2 spaces. It's hard to look at code that switches between indentation styles.

include/libfreenect2/cuda_registration.h
```diff
@@ -0,0 +1,162 @@
...
+   * @param undistorted Undistorted depth frame from apply().
+   * @param registered Registered color frame from apply().
+   * @param[out] cloud_data <X, Y, Z, RGB> coordinates of the 3-D point (meter) and color (BGRX).
+   *     To unpack the color data, use
+   *     const uint8_t *p = reinterpret_cast<uint8_t*>(&rgb);
+   *     uint8_t b = p[0];
+   *     uint8_t g = p[1];
+   *     uint8_t r = p[2];
+   */
+  void getPointXYZRGB(const Frame* undistorted, const Frame* registered, thrust::device_vector<TupleXYZRGB>& cloud_data) const;
```
xlz: This method is mostly for testing. I think most of the users would generate the point cloud with http://wiki.ros.org/depth_image_proc.

So don't change its signature. Add another method to do point cloud generation.

xlz:
There are more issues related to API design. But let's test the algorithm's correctness first.

(But you still need to remove thrust references from build first.) 

The next step is to create test cases that compare CPU and CUDA registration results, specifically for both use cases:

* `apply(rgb, depth, undistorted, registered, true, bigdepth, color_depth_map)`
* `apply(rgb, depth, undistorted, registered, false, bigdepth, color_depth_map)`

The output data structures `undistorted`, `registered`, `bigdepth`, `color_depth_map` should be the same, within floating error.

Please create a branch in your own repo and directly modify Protonect.cpp to facilitate the test cases. I will use that branch to verify the results.

You don't need to worry about API design. I will move them around. There's more to it (I'm thinking changing `Frame` to allow optionally avoiding passing data back to CPU for registration.)

include/libfreenect2/cuda_registration.h
blackzafiro: Thank you for all your comments.  I will address them as soon as I get a chance.

PR #859, opened 2017-06-02
fix the cmake conditional expression error

BlazingForests:


xlz:
Can you explain the difference between `EQUAL` and `STREQUAL`?

BlazingForests:
According to https://cmake.org/cmake/help/v3.0/command/if.html

> if(<variable|string> EQUAL <variable|string>)
>     True if the given string or variable’s value is `a valid number and equal to` that on the right.
>
> if(<variable|string> STREQUAL <variable|string>)
>     True if the given string or variable’s value is `lexicographically equal to` the string or variable on the right.

The difference is `a valid number and equal to` and `lexicographically equal to`

`yes` is not a valid number. 
So even `set(HAVE_CXX11 yes)` the result of  `HAVE_CXX11 EQUAL yes` is `false`

PR #882, opened 2017-07-18
Fixed missing include

ttoocs:
Wasnt compiling, (error: 'bind' is not a member of 'std'). Made it compile.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -159,3 +159,7 @@ $RECYCLE.BIN/
 
 # Mac desktop service store files
 .DS_Store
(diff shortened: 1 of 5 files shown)


xlz:
The correct fix is in frame_listener_impl.cpp
```cpp
#ifdef LIBFREENECT2_THREADING_STDLIB
#include <functional>
#endif
```

I won't accept the other changes.

xlz:
You didn't seem to follow my advice, so I won't accept this PR.

PR #900, opened 2017-08-28
Fixes memory leak in VT processor

fran6co:
When listener doesn't take ownership it will leak the frame


xlz:
It is pretty bad we have to work ourselves to this stuff instead of using unique_ptr. If I were to rewrite libfreenect2 I'd make everything C++11 with a C wrapper.

PR #914, opened 2017-10-05
Add compatibility for L4T 28.1, added make install for udev rules

sjdrc:


xlz:
Please remove the udev install commit. It doesn't work across platforms. It doesn't work across distros even.

xlz:
I cherry-picked the first commit in 2d5cfcf2886c1bac29078cef72bc60a684833ae1.

PR #934, opened 2017-11-24
fix memory leak in DumpDepthPacketProcessor

T0ny0:
Memory leaks if we use `DumpDepthPacketProcessor` and make
```
dev->start()
dev->stop()
```
more then once.
Because every time we call `Freenect2DeviceImpl::startStreams` new tables are created, and old ones are never destroyed.

### DIFF
--- a/src/depth_packet_processor.cpp
+++ b/src/depth_packet_processor.cpp
@@ -159,19 +159,23 @@ const float* DumpDepthPacketProcessor::getZTable() { return ztable_; }
 const short* DumpDepthPacketProcessor::getLookupTable() { return lut_; }
 
 void DumpDepthPacketProcessor::loadP0TablesFromCommandResponse(unsigned char* buffer, size_t buffer_length) {
+  delete[] p0table_;
   p0table_ = new unsigned char[buffer_length];
   std::memcpy(p0table_, buffer, buffer_length);
 }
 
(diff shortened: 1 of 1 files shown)


xlz:
Thanks for spotting the bug.

There is a problem. `loadP0TablesFromCommandResponse` should be able to handle any size and not report errors, according to this usage.

https://github.com/OpenKinect/libfreenect2/blob/6a63380ece335e07584ed55fcf7a0dd8b7fedf41/src/libfreenect2.cpp#L733 

I think adding `delete[]` before `new[]` should be a complete fix (`delete[] nullptr` is okay). What do you think?
```
delete[] p0table_;
p0table_ = new unsigned char[buffer_length];
```

T0ny0:
There is a drawback of
```
delete[] p0table_;
p0table_ = new unsigned char[buffer_length];
```
solution.
We have `const unsigned char* DumpDepthPacketProcessor::getP0Tables() { return p0table_; }` method, and if we take this pointer once and then stop/start device we have a broken pointer.
But now I realized that before my commit it worked the same way, so it will not break someones working code. Fixed the code according to your suggestions.

xlz:
Please also apply this change to `xtable_` and the rest of the relevant variables. Also please squash your changes into a single commit. (`git rebase -i upstream/master; vim: squash, squash...; git push --force`)

PR #602, opened 2016-03-06
Added options to Protonect - [UDP Streamer] and [Image recording to Disk]

PyrApple:
If of use, the merge will probably involve a small restructuration / rewriting for compliance with project's conventions.

PR #942, opened 2017-12-06
Freenect2Replay continued

xlz:
@smokhov I applied some coding style fixup directly. For your review.

TODO

- How to load p0 tables (Read `prefix_123_001.p0table`?)
- How to replay: realtime by timestamp? There should be no loss but the present logic will have packet loss so more work is needed. (replace ready() to synchronous calls)

### DIFF
--- a/include/libfreenect2/libfreenect2.hpp
+++ b/include/libfreenect2/libfreenect2.hpp
@@ -275,17 +275,12 @@ private:
   Freenect2& operator=(const Freenect2&);
 };
 
-class Freenect2ReplayDevice;
+class Freenect2ReplayImpl;
 
 /**
  * Library context to create and open replay devices.
  *
- * You openDevice() and control the device with returned Freenect2ReplayDevice object.
- *
- * You may open devices with custom PacketPipeline.
- * After passing a PacketPipeline object to libfreenect2 do not use or free the object,
- * libfreenect2 will take care. If openDevice() fails the PacketPipeline object will get
- * deleted. A new PacketPipeline object has to be created each time a device is opened.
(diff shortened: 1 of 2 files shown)


smokhov:
@xlz thank you for your work on this

- I see you really made it match a lot more with regular device and its API by doing Freenect2ReplayImpl, etc. Given that do they (Freenect2 and Freenect2Replay) deserve now a common parent since they are so a like? I could use that parent in the other PR in the tool that uses the device; if not it is also OK, can if them out
- You got rid of case insensitivity of the filename extensions, which I suppose is OK if you don't want locale in the lib
- Can also now close #555 as it is also superseded by this work.
- I don't have enough info to comment on the p0 tables TODO
- For replay testing I am working on the app I will be pushing to the other PR to include -replay option to test; it would be convenient to have this one merged

The rest seems like perfect as you know what you are doing :)

xlz:
Sharing a common parent class - they could. But there isn't much duplication right now so extra abstraction isn't as necessary as it appears.

Case sensitivity - yes, I heard locales is a huge mess, https://www.reddit.com/r/programming/comments/7cfftq/wm4_talks_about_c_locales/. The other way is std::transform(..., ::tolower) and http://en.cppreference.com/w/cpp/string/byte/tolower tell us the only way to not have undefined behavior with ::tolower is to use C++11 lambda, uh oh.

TODOs - I was making comments. I thought replay logic will be tricky. How do you synchronize RGB with depth replay orders, how do you handle cross-thread synchronous data flow, etc etc.

smokhov:
@xlz I think your TODOs and thoughts are good candidates for an open issue to avoid being forgotten. RGB is yet to be completed in another PR as well. Thank you for merging. See you in #602!

PR #949, opened 2017-12-20
Are you earthdarkness from the gba

juanpablo64:
Are you this person??? http://gbatemp.net/members/earthdarkness.366355/

if so, can you upload hajimari no ishi tools here?? please¿


floe:
What does this have to do with libfreenect2?

PR #995, opened 2018-05-16
Add support for modern target-based CMake linking

stotko:
This PR adds support for modern target-based linking to freenect2. The old fashioned way of linking, i.e. using `${freenect2_INCLUDE_DIRS}` and `${freenect2_LIBRARIES}`, is preserved for backwards compatibility.


xlz:
To give you some feedback I'm in general for updating CMake to more recent usage. But I can't review this because I don't know what real thing it offers to end-user and I don't know what difference "modern CMake" makes compared to old CMake. CMake as a config language is in no way modern compared to bazel or gyp/ninja. Someone more familiar with the latest development of CMake can review this.

stotko:
The key contribution of this PR is to provide a more robust way of linking against libfreenect2. At the moment, users are forced to use the old error-prone way of linking:
```
target_include_directories(Foo ${freenect2_INCLUDE_DIRS})
target_link_libraries(Foo ${freenect2_LIBRARIES})
```
If one of these variable is misspelled by accident, there will be compiler/linker errors at build time which might be confusing. The modern way using targets, which is added by this PR, not only simplifies linking but is also much more robust:
```
target_link_libraries(Foo freenect2)
```
Misspelling now leads to an error at configure time where CMake complains that the misspelled target does not exist. Furtheremore, the include directories are automatically propagated to the user's target.

More details about modern CMake programming can be found in e.g.:
https://www.slideshare.net/DanielPfeifer1/cmake-48475415
https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/

xlz:
I have since learned about modern CMake, and I agree this is the way to go. But for a maintenance mode project like this such change may incur more potential breakage (mostly third parties depending on particular build interfaces) than we can afford. I will stick to modern CMake best practices in new projects.

stotko:
Thank you for taking a look again at this PR. I agree that there is always at least a minimal risk for breakage. However, I took great care to preserve the old behavior such that a user can upgrade without the need to adjust their code. So this should not cause any regressions.

In case I could still not convince you to merge this PR, I will keep the changes at my fork so that anyone who stumbles over this problem may have the chance to benefit from my unofficial version.

PR #1003, opened 2018-06-18
Pull forward build arguments hint required by downstream libs

eric-schleicher:
Documentation improvement : Added an alternative build for the cx11 options if user are gonig to use with kinect bridge. Hopefully save some people headache/time having to come back to recompile...

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,13 @@ Note: Ubuntu 12.04 is too old to support. Debian jessie may also be too old, and
     make
     make install
     ```
+    **OR** (if you plan to use the[ Kinect2_Bridge](https://github.com/code-iai/iai_kinect2/blob/master/README.md#install))
+    ```
+    mkdir build && cd build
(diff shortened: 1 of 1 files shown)


xlz:
This piece of information is already included in https://github.com/code-iai/iai_kinect2/blob/master/README.md#install, no?

eric-schleicher:
correct... it is, but having it here (as an alternative) mean people can see it and won't have to get to kinect)bridge documentation, and **then** come back to recompile when(if) they find it there.   Not a big deal, I thought it was a small improvement, having been caught out by it previously.

whenever i was creating a new system, I had taken to opening the kinect2_bridge documentation first to get the value (prior to memorizing it), this would have just made it easier.

xlz:
Downstream projects can make whatever changes based on their needs (kinect2 bridges also recommends `-DCUDA_PROPAGATE_HOST_FLAGS=off`, what else?). We can't be responsible for all that.

PR #1028, opened 2018-10-10
add libusb install script for vs2017 community

xxzefgh:


xlz:
libusb has released a new version with all the needed features: https://github.com/libusb/libusb/releases/tag/v1.0.22. Can you see if you can use the official build libusb-1.0.22.7z and update README instead?

xlz:
Fixed in https://github.com/OpenKinect/libfreenect2/commit/73e1a24a2f3b28bd35eed9778dc680883bf9190f.

PR #1033, opened 2018-10-30
Add exposure control of the color camera

### LINKED ISSUE #1031: Adding exposure control for Kinect2
Has anyone investigated adding manual exposure control to libfreenect2?

I know it is possible to manually control exposure of the Kinect2 by using a library that Microsoft included in a project they released on GitHub: https://github.com/Microsoft/MixedRealityCompanionKit/tree/master/KinectIPD/NuiSensor

This library, "NuiSensorLib", is a precompiled binary for Windows. I currently use it with a small C++ wrapper to set the frame integration time and the analog gain of the Kinect2 during recording. The relevant library function is NuiSensor_ColorChangeCameraSettings, and the other files beside this library file have examples of usage.

I would like this functionality within libfreenect2 so I can control multiple Kinect2 devices on a single PC.

The code I use constructs a command message with a sequence number of zero, and an RGB setting command integer and a value integer. The problem is that I suspect the NuiSensor_ColorChangeCameraSettings function then wraps this in another command structure before sending over USB, and since it is a pre-compiled library, I don't have the values that it is using.

Has anyone else looked into this library already? Or have any hints on where I could look next?

Thanks 😄

saulthu:
fixes #1031


floe:
Excellent, thank you very much. I don't have a device for testing right now, but the code looks very reasonable to me. Could you perhaps add a reference to the NuiSensor.h file where the information originally came from? Since that's under MIT license, and your code is not a verbatim copy of any parts, I think this should be good to merge, at least license-wise.

saulthu:
You're most welcome, and thanks for all of your work on this :). I added `NuiSensorLib.h` (with a prefix of the original MIT license) to the directory `include/internal/libfreenect2/protocol` since it contains a lot of hints for protocol and structures that pass over the wire.

include/internal/libfreenect2/protocol/NuiSensorLib.h
```diff
@@ -0,0 +1,1356 @@
+//--------------------------------------------------------------------------
```
xlz: Why is this file needed in this repo? I looked around and it doesn't seem to be used anywhere. A reference to https://github.com/Microsoft/MixedRealityCompanionKit/blob/master/KinectIPD/NuiSensor/NuiSensorLib.h would suffice.

src/libfreenect2.cpp
```diff
@@ -736,6 +743,57 @@ void Freenect2DeviceImpl::setIrAndDepthFrameListener(libfreenect2::FrameListener
...
+  CommandTransaction::Result result;
+  command_tx_.execute(ColorSettingCommand((ColorSettingCommand::COMMAND_TYPE)cmd), result);
+  ColorSettingResponse const* data = reinterpret_cast<ColorSettingResponse const*>(&result[0]);
+  return data->Data;
+}
+
+float Freenect2DeviceImpl::getColorSettingFloat(uint32_t cmd)
+{
+  uint32_t data = getColorSetting(cmd);
+  return *reinterpret_cast<float const*>(reinterpret_cast<uint8_t const*>(&data));
```
xlz: As before, use
```
float out;
memcpy(&out, &data, sizeof(out));
return out;
```

src/libfreenect2.cpp
```diff
@@ -736,6 +743,57 @@ void Freenect2DeviceImpl::setIrAndDepthFrameListener(libfreenect2::FrameListener
...
+{
+  CommandTransaction::Result result;
+  command_tx_.execute(ColorSettingCommand((ColorSettingCommand::COMMAND_TYPE)cmd, value), result);
+}
+
+uint32_t Freenect2DeviceImpl::getColorSetting(uint32_t cmd)
+{
+  CommandTransaction::Result result;
+  command_tx_.execute(ColorSettingCommand((ColorSettingCommand::COMMAND_TYPE)cmd), result);
+  ColorSettingResponse const* data = reinterpret_cast<ColorSettingResponse const*>(&result[0]);
```
xlz: Must check if result.size() > sizeof(ColorSettingResponse) before doing the cast here.

include/internal/libfreenect2/protocol/NuiSensorLib.h
saulthu: I misunderstood the comment above to add a reference, so I've now removed the header.

include/libfreenect2/libfreenect2.hpp
```diff
@@ -160,6 +160,45 @@ class LIBFREENECT2_API Freenect2Device
...
+  virtual void setColorSemiAutoExposure(float pseudo_exposure_time_ms) {}
+
+  /** Manually set true exposure time and analog gain of the RGB camera.
+   * @param integration_time_ms True shutter time in milliseconds, range (0.0, 66.0]
+   * @param analog_gain Analog gain, range [1.0, 4.0]
+   */
+  virtual void setColorManualExposure(float integration_time_ms, float analog_gain) {}
+
+  /** Set/get an individual setting value of the RGB camera. */
+  virtual void setColorSetting(uint32_t cmd, uint32_t value) {}
```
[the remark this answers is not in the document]
saulthu: These additional functions are not necessary for the exposure controls. Will remove them.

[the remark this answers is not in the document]
saulthu: On second thought, I'll move it to a public header.

saulthu:
Thanks for the review :). I made commits to address all review comments.

xlz:
The only question is, does it work in testing?

saulthu:
I have tried it within a modified protonect build on Win10.

The original code that I used, which is quite similar, has been in use for some time now, and this is a port of that code, specifically the usage of setting exposure as auto v's semi-auto v's manual which is not documented in the released MS library.

Is there a place where I should put specific tests? Or should I add an option to protonect?

xlz:
Can you consolidate the commits into a single one (rebase with fixup)? Looks like I can't squash and merge while keeping your name.

saulthu:
OK, all commits are fixup'd into a single commit.

PR #1052, opened 2019-01-06
Add setting of LEDs

saulthu:
This allows setting of LEDs on the Kinect2 sensor. The LEDs can be set independently to various brightness levels [0, 1000], and also to blink between two brightness levels at some frequency (by giving a half-period value in milliseconds). The two LEDs can have different settings.

This is another product of my debugging though Microsoft's [NuiSensorLib](https://github.com/Microsoft/MixedRealityCompanionKit/tree/master/KinectIPD/NuiSensor). The NuiSensorLib library only ever sets the LEDs to a constant brightness, either on or off, but debugging shows the underlying structure with extra fields for LED mode (on or blink) and blink mode plus time between toggle.

### DIFF
--- a/src/libfreenect2.cpp
+++ b/src/libfreenect2.cpp
@@ -263,6 +263,7 @@ public:
   virtual void setColorSetting(ColorSettingCommandType cmd, float value);
   virtual uint32_t getColorSetting(ColorSettingCommandType cmd);
   virtual float getColorSettingFloat(ColorSettingCommandType cmd);
+  virtual void setLedStatus(LedSettings led);
   virtual bool start();
   virtual bool startStreams(bool rgb, bool depth);
   virtual bool stop();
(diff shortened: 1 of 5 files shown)


saulthu:
Fixed some typos and class name.

saulthu:
I make use of the blinking mode, which allows getting user attention more effectively than just turning on. It's much nicer than having to repeatedly set on/off manually.

src/libfreenect2.cpp
```diff
@@ -810,6 +812,12 @@ float Freenect2DeviceImpl::getColorSettingFloat(ColorSettingCommandType cmd)
   return out;
 }
 
+void Freenect2DeviceImpl::setLedStatus(LedSettings led)
+{
+  CommandTransaction::Result result;
+  command_tx_.execute(LedSettingCommand(led), result);
```
xlz: Do you need any validation here? Can invalid values damage the hardware?

saulthu: Invalid values appear to be dropped e.g. a LED brightness outside of [0, 1000] e.g. 1001 results in the LED being turned off. So, the Kinect2 is doing brightness validation. The frequency is probably implemented with a hardware timer, and is very unlikely to cause problems for blinking. I've tried many values and all my Kinects still work.

PR #1060, opened 2019-02-10
Add build script for Android

CismonX:
Add a simple NDK build script for Android support.
Tested on my Hikey 970 board running Android P.

Future support:
* Java bindings.
* OpenGL depth packet processor.

### DIFF
--- a/platform/android/README.md
+++ b/platform/android/README.md
@@ -1,0 +1,42 @@
+# Build libfreenect2 for Android
+
+## 1. Requirements
+
+* A latest release of Android NDK.
+* A rooted Android device with USB3.0 host.
+
+## 2. Build libusb
+
+* See the [official documentation](https://github.com/libusb/libusb/blob/master/android/README) for instructions to build libusb for Android.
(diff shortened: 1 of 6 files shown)


xlz:
I can merge most of this save for the last bit where you make TurboJPEG optional. It's not optional because you can't dump RGB packets when it's not available. Dumping RGB packets makes the essential function of this library not working.

CismonX:
> I can merge most of this save for the last bit where you make TurboJPEG optional. It's not optional because you can't dump RGB packets when it's not available. Dumping RGB packets makes the essential function of this library not working.

I believe there are cases when you just want to use raw RGB or depth frames without first rendering them to displayable image, such as storing them to files for later processing. Real-time packet processing yields low frame rate on Android devices since many packets have to be discarded due to poor hardware performance.

xlz:
You're changing the default to a dysfunctional configuration. If realtime processing is slow you can opt in for the packet dumper. It's a different question of changing the default.

CismonX:
> You're changing the default to a dysfunctional configuration. If realtime processing is slow you can opt in for the packet dumper. It's a different question of changing the default.

Okay. I've just removed the second commit from the pull request.

Issue #1082, opened 2019-08-12
TegraJPEG support for recent L4T versions (e.g. 28.3.1)


dbruning:
I'm using a Jetson Nano and trying to find out how quickly it can take frames from the Kinect v2 camera.

Out of the box, it's very slow with the RBG frames, so I'm trying to enable TegraJPEG support.
It seems that cmake_modules/FindTegraJPEG.cmake is responsible for downloading and compiling TegraJPEG library, but there are a couple of problems with it:

1. it tries to find l4t version by parsing /etc/nv_tegra_release, but that file [isn't there anymore](https://devtalk.nvidia.com/default/topic/1057675/jetson-nano/read-l4t-version-from-jetpack-4-2-1/)

2. I tried a workaround of guessing the URL of the BSP source, based on the previous examples in that .cmake file. My guess was http://developer.download.nvidia.com/embedded/L4T/r28_Release_v3.1/BSP/source_release.tbz2 (because 28.3.1 seems to be the latest release of L4t) but that link isn't correct. Perhaps the location of those sources changed recently.

Does anyone know the current location of the BSP sources for the Nano? 

Disclaimer: I'm new to C and CMake so apologies if I've made an obvious errors.

xlz:
The CMake file is trying many ways to find Nvidia's custom `jpeglib.h`. In a recent installation it looks like this file is installed as part of "Tegra Multimedia API" in /usr/src/tegra... but I'm not sure all people include it during installation.

Anyway, that CMake file is only supposed to produce the INCLUDE PATH of `jpeglib.h` and LINKING `-lnvjpeg`. Maybe you can manually add the two flags somewhere else.

dbruning:
Thanks for the reply & the advice. I've managed to get it working (very messily) in my fork [here](https://github.com/dbruning/libfreenect2/commit/4929cb82651ede350d772790a41076ad0d81810e). 
It runs with a good framerate pulling both RGB and depth frames (if I disable the registration code), but will crash after about 5 seconds, having used up all the memory on the Nano. Next time I get a chance to work on this I'll try to find the memory leak - perhaps something has changed in the Nvidia JPEG library since it was first used by libfreenect2.
If I can get it all working properly I'll tidy it up & make a pull request.

chunter1:
@dbruning I got the same probem. Could you fix the crashing?

dbruning:
[...]

Issue #774, opened 2016-12-22
Wrong colours when using TegraJPEG for RGB processing on Jetson TK1


ksze:
Overview Description:

Version, Platform, and Hardware Bug Found:

1. `git log -1 --oneline`
    cec93c7 Merge pull request #765 from xlz/disable-default-constructors

2. `uname -a`

        Linux tegra-ubuntu 3.10.40-ga7da876 #1 SMP PREEMPT Tue Jun 7 21:05:06 PDT 2016 armv7l armv7l armv7l GNU/Linux`

3. `lsusb -t`

        /:  Bus 03.Port 1: Dev 1, Class=root_hub, Driver=tegra-ehci/1p, 480M
            |__ Port 1: Dev 5, If 0, Class=Hub, Driver=hub/4p, 480M
                |__ Port 1: Dev 6, If 0, Class=Human Interface Device, Driver=usbhid, 1.5M
                |__ Port 2: Dev 7, If 0, Class=Human Interface Device, Driver=usbhid, 1.5M
        /:  Bus 02.Port 1: Dev 1, Class=root_hub, Driver=tegra-xhci/2p, 5000M
            |__ Port 1: Dev 4, If 0, Class=Hub, Driver=hub/1p, 5000M
                |__ Port 1: Dev 5, If 0, Class=Vendor Specific Class, Driver=, 5000M
                |__ Port 1: Dev 5, If 1, Class=Vendor Specific Class, Driver=, 5000M
                |__ Port 1: Dev 5, If 2, Class=Audio, Driver=snd-usb-audio, 5000M
                |__ Port 1: Dev 5, If 3, Class=Audio, Driver=snd-usb-audio, 5000M
        /:  Bus 01.Port 1: Dev 1, Class=root_hub, Driver=tegra-xhci/6p, 480M
            |__ Port 3: Dev 3, If 0, Class=Hub, Driver=hub/1p, 480M
4. `lspci -nn`

        00:00.0 PCI bridge [0604]: NVIDIA Corporation TegraK1 PCIe x1 Bridge [10de:0e13] (rev a1)
        01:00.0 Ethernet controller [0200]: Realtek Semiconductor Co., Ltd. RTL8111/8168/8411 PCI Express Gigabit Ethernet Controller [10ec:8168] (rev 0c)

Steps to Reproduce:

1. Configure libfreenect2 with `ENABLE_TEGRAJPEG` in cmake;
2. Build it.
3. Run `Protonect cudakde`

[...]

PR #1107, opened 2020-02-17
Fix color unmatch when TegraJpeg is enabled for both Protonect and OpenNI2 Driver #774

yjmade:
TegraJPEG only support return RGB but libfreenect2 suppose all jpeg result is BGR.  #774 #1082

### DIFF
--- a/cmake_modules/FindTegraJPEG.cmake
+++ b/cmake_modules/FindTegraJPEG.cmake
@@ -6,7 +6,7 @@
 FIND_PATH(TegraJPEG_INCLUDE_DIRS
   libjpeg-8b/jpeglib.h
   DOC "Found TegraJPEG include directory"
-  PATHS /usr/src/tegra_multimedia_api/include
+  PATHS /usr/src/tegra_multimedia_api/include /usr/src/jetson_multimedia_api/include
   NO_DEFAULT_PATH
 )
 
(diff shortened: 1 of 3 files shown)


xlz:
libfreenect2 does not suppose all jpeg result is BGR. It merely passes through what is decoded from jpeg. This introduces additional processing which may not be needed by the user. The viewer showing wrong colors is not a problem.

yjmade:
> libfreenect2 does not suppose all jpeg result is BGR. 

At least the OpenNi2 driver part expect the color stream input would be in BGR, and it store the data as RGB. 

> `modes[makeOniVideoMode(ONI_PIXEL_FORMAT_RGB888, 1920, 1080, 30)] = 1;`

And the origin code of `ColorStream::copyFrame` actually did flipping the color. 

What I did is just make this method use the origin order of color from decoder, in this case is RGB, to match what OpenNI2 expect. This will not introduce addition computation.

xlz:
OpenNI2 says it expects BGR but does it do anything that using RGB in place of BGR would make a difference? My expectation is that this RGB be passed down the pipeline as false BGR until where it is to be used as actual BGR and until then would this RGB be converted to BGR, not using this CPU swapping, you know, but with Tegra specific hardware accelerated routines that have to be provided by the user as it's inconvenient for this library to introduce more dependency on platform specific APIs.

PR #1163, opened 2022-01-01
Release 0.2.1 is NOW compatible with OpenCV 4.

jiapei100:
```CV_IMWRITE_JPEG_QUALITY -> cv::IMWRITE_JPEG_QUALITY```


smokhov:
@jiapei100 -- is it still backward compatible with OpenCV3? I still have apps that need that version.

